Issue #9, opened 2021-11-30
Integrate with profile


TheTechArch:
## Description
We need to integrate with profile to get the addresses from KRR when sender does not have them

## Development task
- [x] Create client to call profile. 
- [x] Configure access token usage from notification to profile (Partially completed)
- [x] Add code to find address when not provided from sender (Moved to new issue)

The last task is moved to a new issue: #22

SandGrainOne:
We've implemented a client to get a UserProfile from the Profile component. One method with UserId and another with "SSN" - National identity number. 

The Client will require an IAccessTokenGenerator to get an AccessToken and an IUserTokenProvider to get the user token before submitting the request. Both have implementations, but neither has been added to the ServiceProvider. Some additional configuration might be needed.

The test coverage in the Integration project is currently 74%. On new code it is close to 90%

PR #20, opened 2021-12-07
Adding integration with the Profile component

SandGrainOne:
Adding a typed http client for Notification to send requests to the Profile component.

#9


src/Altinn.Notifications.Integrations/IUserTokenProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Altinn.Notifications.Integrations
+{
+    /// <summary>
+    /// Defines the methods required for an implementation of a user JSN Web Ttoken provider.
```
acn-sbuad, line 4:
```suggestion
    /// Defines the methods required for an implementation of a user JSN Web Token provider.
```

src/Altinn.Notifications.Integrations/PlatformSettings.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the Profile component API base url.
+        /// </summary>
+        public string ProfileEndpointAddress { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Gets or sets the name of the subscription key request header.
+        /// </summary>
+        public string ProfileSubscriptionKeyHeaderName { get; set; } = "Ocp-Apim-Subscription-Key";
```
acn-sbuad, line 16: Could we give this property a more generic name? I guess we will need the same header for requests towards register shortly.

SandGrainOne: It's technically possible for API Management to have a different header name for each API. Though I don't see why we would ever use that I kind of like having support for it in the client.

Issue #13, opened 2021-12-06
Update send state for successfully sent target


acn-sbuad:
## Description

We need a way to identify which targets have been successfully (and unseccessfully) sent. 

## Development task
- [ ] Create function for updating sent for a target
- [ ] Update function for retrieving unsent targets
- [ ] Implement code for call to update target once successfully sent

SandGrainOne:
No longer relevant.

PR #21, opened 2021-12-07
Feature/13 mark target as sent

acn-sbuad:
#13


src/Altinn.Notifications/Migration/v0.02/01-setup-functions.sql
```diff
@@ -0,0 +1,42 @@
...
+        "address" character varying,
+        "sent" timestamptz)
+    LANGUAGE 'plpgsql'
+
+AS $BODY$
+BEGIN
+return query
+	SELECT t.id, t.notificationid, t.channeltype, t."address", t."sent"
+    WHERE t."sent" IS NULL
+	FROM notifications.targets t;
```
SandGrainOne, line 16: We usually place the FROM clause before the WHERE clause. Curious that this even works.

tba76 (approved):
🍺

PR #27, opened 2022-01-03
Some code smells removed, other simply moved.

SandGrainOne:
Fixed a few code smells. In many cases related to null references the changes simply moves the issue to other method calls, but it is a step in the right direction.


tjololo (commented):
I see the test/Altinn.Notifications.Tests/Mocks/NotificationRepositoryMock.cs does not match the implemented member. If we relay on this to test the code it should at least have the same return type

tjololo (approved):
Steps in the right direction

PR #28, opened 2022-01-11
Feature/17 k8s deploy workflow

### LINKED ISSUE #17: Set up workflow for deploy to k8s cluster
Prerequisite: figure out how to authenticate the workflow againt the cluster

Set up deploy. 
Should first be deployed to AT. 
Should not use the latest tag, so we can control the version in the environment at all times

tjololo:
Automate deploy to at22 with github actions
Fixes #17


acn-sbuad (approved):
🎈

PR #31, opened 2022-01-17
Fixing a few compile warnings related to nullability

SandGrainOne:


src/Altinn.Notifications.Persistence/NotificationRepository.cs
```diff
@@ -233,9 +233,9 @@ public async Task<List<Target>> GetUnsentTargets()
             return target;
         }
 
-        public async Task<Target> UpdateSentTarget(int id)
+        public async Task<Target?> UpdateSentTarget(int id)
```
RonnyB71, line 236: Must admit this was a bit confusing. A method named Update... that runs a select - or is the notifications.update_senttarget(@id) updating something?

src/Altinn.Notifications.Core/INotificationsRepository.cs
```diff
@@ -10,7 +10,7 @@ public interface INotificationsRepository
 
         Task<Target> AddTarget(Target target);
 
-        Task<Target> UpdateSentTarget(int id);
+        Task<Target?> UpdateSentTarget(int id);
```
RonnyB71, line 13: Why introduce a nullable?

SandGrainOne: I think it is the best way of transferring the knowledge that the target doesn't exist to the caller.

src/Altinn.Notifications.Persistence/NotificationRepository.cs
SandGrainOne: I'm not overly happy about the structure of the SQL command text either. There might be a better way, but I'm not too familiar with the quirks of PostgreSQL. 

The command creates a call to a function that updates the target and returns a row that is selected.

src/Altinn.Notifications.Core/INotificationsRepository.cs
RonnyB71: But when we pass in an id, then it sounds more like an exception than that target does exists - I assume it must have existed at some point since we have an id - or am I missing something from the bigger picture?

src/Altinn.Notifications.Persistence/NotificationRepository.cs
RonnyB71: Any particular reason why we use functions in the database and not in code? Or at least a call to an update statement instead of a function?

SandGrainOne: I believe it provides better protection against SQL injection. When PostgreSQL handles the transfer of values through function parameters.

RonnyB71: I would assume any sql statement could be parameterized to prevent sql injection including an simple update statement. Also in this case it's an int that's passed in, so don't really see why we need the extra layer of complexity by "hiding" parts of the code in an SQL function.

src/Altinn.Notifications.Core/INotificationsRepository.cs
RonnyB71: If this is used as signaling to the controller that the expected id passed in wasn't found then I would consider throwing an exception at this level and catch that in the controller with the reasoning that it is expected that the id is valid and should be there, it's really an exceptional case that it's not there.

src/Altinn.Notifications.Persistence/NotificationRepository.cs
SandGrainOne: This is a fine discussion to have as a team.

src/Altinn.Notifications.Core/INotificationsRepository.cs
SandGrainOne: I'm not sure if Repository should assume anything of how it's methods are used or the context of the calls. I did consider letting the reader logic fail with an exception if there were no rows to read, but it didn't feel right in isolation.

RonnyB71: > I'm not sure if Repository should assume anything of how it's methods are used or the context of the calls.
Agree. That's basically why I'm proposing to throw an exception, because if an id is missing then something is wrong, unless records are wiped out on a regular basis. Then returning null makes more sense, or perhaps even a null object. Returning null forces a null check to be done on the caller side if you expect the record to be there in all/most situations. You said almost all cases, what are the cases when we don't expect to find an id?

> I did consider letting the reader logic fail with an exception if there were no rows to read, but it didn't feel right in isolation.
Why didn't it feel right?

Sorry for nagging on about this, but I quite don't get the case - and an id that earlier has been confirmed shouldn't just wannish.

SandGrainOne: Current logic makes no assumption about whether the target exists. This is in line with the old code with the difference that the old code produced compile time warnings. The end result was the same.

I'm not entierly oposed to changing it. That was my initial idea. I went for the other way mostly because I was thinking that this is the best way to handle the missing target case.

RonnyB71 (approved):
🕹️

Issue #164, opened 2023-05-11
Endpoint for Posting a notification order (publish event for immediate sending)


tba76:
## Description

Endpoint for posting a notification (syncron) should save to database and push to Kafka-topic.
## Considerations

- Is from address required to provide? What is our default sender? 
-  Is sendersReference required? 
- 
## Specification tasks
- [swagger for  API  ](https://app.swaggerhub.com/apis/STEPHANIEBUADU/Notifications/1.0.0#/EmailNotificationOrderRequest)
## Development tasks
- [x] Create new controller and endpoint for post notification
- [x] Create service for save notification
- [x] Create necessary database tables (procedures, functions?)
- [x] Create repository-classes
~~- [ ] Add authorization to endpoint (ScopeOrAccessToken)~~ should be handled in #8
- [x] Documentation
  - [x] Swagger
- [x] QA
- [x] Manual test is complete (if relevant)

## Acceptance criteria
- [x] Posting a new notification should save to postgres 
- [x] Invalid input should be denied (bad request)
- [x] Unauthorized users should be denied post

SandGrainOne:
[openapi3_1.yaml.txt](https://github.com/Altinn/altinn-notifications/files/11881876/openapi3_1.yaml.txt)

PR #184, opened 2023-07-13
Logic for registering new email notification request

acn-sbuad:
### Implemented considerations
- [x] Created services for DateTime and Guid to facilitate unit testing.
  - [ ] Alt1:  We could alternatively handle it by creating custom Assert methods for the complex objects to exclude id and datetime from comparison. 
  - [ ] Alt2: We could use fluentAssertion to make use of the `exclude` functionality

- [ ] Internal empty constructors for all domain models. Ensures we can easily create objects for testing and within service logic, but forcing other projects to use official constructor. 

- [x] Which statues do we use to follow state for an order? When do we know we need to reprocess a particular order

- [ ] NotificationOrderRequest vs NotificationOrder vs EmailNotificationOrderRequestExt. Det blir mange modeller, men ser at dette kan være nyttig for å skille mellom ting. Alternativet må være at controller mapper til notificationOrder, men da må kontroller kjenne til intern logikk eller opprette et uferdig object som sendes videre til service. Det er mindre ønskelig. 
  - [ ] NotificationOrder (en bestilling som gjelder én melding som skal kommuniserer til mottaker(e) vi én eller flere kanaler) Eies av domene laget.
  - [ ]  NotificationOrderRequest. Input til service. Én NotificationOrderRequest kan resultere i flere notificationOders, f.eks hvis man har med info om revarsling og krav for dette. 
  - [ ] EmailNotificationOrderRequestExt. Ekstern modell eies av API-laget. 
  
### Not implemented considerations
~~- [ ] Validation rule ensuring all provided email addresses in notificationOrderRequests are unique.~~
- [ ] Validation to ensure senders reference is unique within the creator scope? 
- [ ] Validation of from address based on creator / org
- [ ] Which indexes do we need in the db? 

## Related Issue(s)
- #164 
## Verification
~~- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)~~
- [x] All tests run green

## Documentation


src/Altinn.Notifications.Persistence/Repository/OrderRepository.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        _dataSource = dataSource;
+    }
+
+    /// <inheritdoc/>
+    public async Task<NotificationOrder> Create(NotificationOrder order)
+    {
+        long dbOrderId = await InsertOrder(order);
+
+        EmailTemplate? emailTempate = ExtractTemplates(order);
```
acn-sbuad, line 35: typo

src/Altinn.Notifications.Core/Enums/AddressType.cs
```diff
@@ -5,5 +5,7 @@
 /// </summary>
 public enum AddressType
 {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
     Email
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
```
tba76, line 10: If we should suppress this warnings in enums, make it identical in every enum.

acn-sbuad: sonarcloud marks it as code smells. We would consider marking it as won't fix in sonarcloud to avoid the supressions as well. What do you think?

tba76: We suppress warnings so Visual Studio is happy :)

tba76 (approved):
🍺 🍺 
  🍻

PR #212, opened 2023-08-10
Fixed failing health checks

acn-sbuad:
## Description

- [x] Only produce messages on kafka topic
- [x] Custom postgres health check implementation


tba76 (approved):
🍺

PR #213, opened 2023-08-11
Cleanup of kafka clients

acn-sbuad:
## Description
- BackgroundService for consumer instead of IHostedService
- Logic for adjusting config based on env
- shared base logic for consumer

Remaining work: 
- make consumer more generic to parse any type

## Verification
- [ ] All tests run green


src/Altinn.Notifications.Integrations/Kafka/Consumers/KafkaConsumerBase.cs
```diff
@@ -0,0 +1,99 @@
...
+using Confluent.Kafka;
+
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Altinn.Notifications.Integrations.Kafka.Consumers;
+
+/// <summary>
+/// Base class for Kafka Consumer of  Notification Orders
```
acn-sbuad, line 13:
```suggestion
/// Base class for Kafka Consumer of Notification Orders
```

src/Altinn.Notifications.Integrations/Kafka/Consumers/KafkaConsumerBase.cs
```diff
@@ -0,0 +1,99 @@
...
+        string message = string.Empty;
+        try
+        {
+            while (!stoppingToken.IsCancellationRequested)
+            {
+                var consumeResult = _consumer.Consume(stoppingToken);
+                if (consumeResult != null)
+                {
+                    message = consumeResult.Message.Value;
+                    bool succeeded = NotificationOrder.TryParse(message, out NotificationOrder? order);
```
acn-sbuad, line 76: future goal would be to make this even more generic so we can state the type in the constructor, and use a json serializer of some sorts. Might also require that we define schemas for our topics to ensure consistent data

PR #225, opened 2023-08-18
set client config based on presence of username and pw

acn-sbuad:
## Description

Changed implementation of shared client config to not depend on environment


tba76 (approved):
🍺

Issue #262, opened 2023-10-10
Adjust procedure for past due orders to retrieve element with sendtime in the near future as well


acn-sbuad:
### Description

We have seen that it takes a bit of time for an email to flow through the whole process. 
So by retrieveing order that have a requested send time that is in near future, we can better hit the requested send time for when the email is actually received by the end user. 

Suggestion is to define near future av 30 sec or less. 

### Tasks
- [x] Modify function for retrieving past due orders to give 30 seconds extra on the requested send time. 
- [x] QA
- [x] Test

### Acceptance Criterias
- [x] An order with requested send time more than 30 seconds in the future should not be picked up by the function 
- [x] An order with requested send time less than 30 seconds away should be picked up by the function

PR #274, opened 2023-10-16
adjust grace period to 1 min

acn-sbuad:
## Description

Edited existing release, will run script manually in AT envs to save number of migration version. 

1 minute makes more sense wr.t. to the schedule of the cron-jobs. 

## Related Issue(s)
- #262


tba76 (approved):
🍺

PR #285, opened 2023-10-24
DB tuning: json convertion and dependency tracking for postgres

HenningNormann:
## Description
- Let npgsql do json convertion for notificationorder
- Added dependency tracking for postgres
- Added loop in StartProcessingPastDueOrders
- (One commit for each)

## Related Issue(s)
- #{253}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.Notifications.Core/Services/OrderProcessingService.cs
+++ b/src/Altinn.Notifications.Core/Services/OrderProcessingService.cs
@@ -53,7 +53,7 @@ public class OrderProcessingService : IOrderProcessingService
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


acn-sbuad:
Could we consider using decorators to achieve the tracking addition instead of adding the AI dependency to the repository classes? Implemented an example for the emailnotificationRepository here to illustrate what it would look like. 
#286 
What we would gain would be that no changes would be required in the repository class, limiting it to only required logic

acn-sbuad (approved):
🎈

PR #307, opened 2023-11-16
removed Xss180Kseting for kafdrop #GCPActive

acn-sbuad:
## Description

Bugfix for failing config after kafdrop upgrade


tomshag (approved):
This fixes the issue that occurs after latest kafka image update.

PR #308, opened 2023-11-22
Validating `requestedSendTime` to check if it has zone specified or not

### LINKED ISSUE #275: Providing a requestedSendTime without zone indicator gives InvalidCastException
### Description of the bug

While testing issue #262 I discovered that providing a time stamp without a timezone indicator produced an internal server error. I get the same error if I indicate timezone different from UTC.

We should give a 400 respons if no timezone is given. 

### Steps To Reproduce

1. Create a new notification order request with a requestedSendTime without timezone or timezone different from UTC.
```
{
    "subject": "Tester utsending av HTML epost",
    "body": "<html><body><h1>My First Heading</h1><p>My first paragraph.</p></body></html>",
    "content-type": "Html",
    "requestedSendTime": "2023-10-17T12:00:00",
    "sendersReference": "testissue262",
    "recipients": [
        {
            "emailAddress": "[REDACTED]"
        }
    ]
}
```
2. Observe the InternalServerError response.

### Additional Information

Should we force the use of UTC if given timezone is wrong or should we do a silent convert?
```[tasklist]
### Tasks
- [x] Add validation to ensure that timezone is specified in the input model
- [x] Add logic to convert local time to UTC before the value is stored, Convert in controller.
```

codemascot:
Validating `requestedSendTime` to check if it has zone specified or not.

## Description
Introduced a method name `HasTimeZone` in the `EmailNotificationOrderRequestValidator` validator and implemented validator with this method. If the zone isn't specified then it'll throw a 400 HTTP error.

## Related Issue(s)
- FIX #275 

## Verification
- [x] All tests run green

### DIFF
--- a/src/Altinn.Notifications/Validators/EmailNotificationOrderRequestValidator.cs
+++ b/src/Altinn.Notifications/Validators/EmailNotificationOrderRequestValidator.cs
@@ -23,7 +23,7 @@ public class EmailNotificationOrderRequestValidator : AbstractValidator<EmailNot
             .WithMessage("A valid email address must be provided for all recipients.");
 
         RuleFor(order => order.RequestedSendTime)
-            .Must(sendTime => HasTimeZone(sendTime))
+            .Must(sendTime => sendTime.Kind != DateTimeKind.Unspecified)
             .WithMessage("No time zone specified.")
             .Must(sendTime => sendTime >= DateTime.UtcNow.AddMinutes(-5))
             .WithMessage("Send time must be in the future. Leave blank to send immediately.");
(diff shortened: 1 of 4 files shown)


src/Altinn.Notifications/Validators/EmailNotificationOrderRequestValidator.cs
```diff
@@ -23,8 +23,10 @@ public EmailNotificationOrderRequestValidator()
             .WithMessage("A valid email address must be provided for all recipients.");
 
         RuleFor(order => order.RequestedSendTime)
-          .Must(sendTime => sendTime >= DateTime.UtcNow.AddMinutes(-5))
-          .WithMessage("Send time must be in the future. Leave blank to send immediately.");
+            .Must(sendTime => sendTime.Kind != DateTimeKind.Unspecified)
+            .WithMessage("No time zone specified.")
```
SandGrainOne, line 27: We could be slightly more specific in the error message. Suggestion:
"The requested send time value must have specified a time zone."

codemascot:
Two tests failed in my Mac local testing environment:
1) `Altinn.Notifications.IntegrationTests.Notifications.Integrations.PostgresHealthCheckTests.CheckHealthAsync_InvalidConnectionString_ReturnsUnhealthyResult`
Here this is expecting `Unhealthy`, while my machine returns `Healthy` which is most probably because my local PostgreSQL is exposed at the `5432` port by default. 🤔 

2) `Altinn.Notifications.Tests.Notifications.Integrations.TestingExtensions.ServiceCollectionExtensionsTests.AddKafkaServices_KafkaSettingsMissing_ThrowsException`

Otherwise all the other tests came green.

acn-sbuad (approved):
🎈

PR #379, opened 2024-01-15
fixed code smells from sonarcloud #GCPActive

acn-sbuad:
- k6 workflows running on docker task. 
- Added regression test workflow to each env. Runs at noon each weekday


tba76 (approved):
🍺

Issue #336, opened 2023-12-08
Core logic for placing an sms notification order


acn-sbuad:
### Description

Workflow: order placement for SMS notifications will need modifications to the Core layer of the notification app that can support the new endpoint. 

### Tasks
- [x] Extend` NotificationOrder.cs` with required properties to represent an SMS template
- [x] Extend `NotificationOrderRequest.cs` with required properties to represent an SMS template
- [x] Rename `EmailNotificationOrderService` and `IEmailNotificationOrderService`  to `OrderRequestService` and `IOrderRequestService`
- [x] Rename `RegisterEmailNotificationOrder` in `OrderRequestService`  to `RegisterNotificationorder`
- [x] Extend SetFromAddressIfNotDefined in OrderRequestService with logic for setting default SMS sender
- [x] Implement unit tests
- [x] QA
~- [ ] Manual test~ Manual testing not possible as no external endpoints are available

### Acceptance Criterias
- [x] Required intefaces are defined and ready to use by API layer
- [x] Required intefaces are defined and ready to use by persistence layer
- [x] Domain models for `NotificationOrder` and `NotificationOrderRequest` have properties to represent SMS notification templates

acn-sbuad:
Regressoin tests ok. If any future issues occur, these will be handled in relation to other /new issues

PR #380, opened 2024-01-16
#336 Implementation of core logic for placing an sms notification order

codemascot:
This PR is made to implement the core logic for placing an sms notification order described on the issue #336 

## Description
- Added SMS channel to `NotificationChannel` enum.
- Added SMS notification template type to the `NotificationTemplateType` enum.
- Added SMS template(`SmsTemplate`) to the `NotificationTemplate` directory.
- Added JSON subtype(`JsonDerivedType`) for SMS in the notification interface.
- Renamed `EmailNotificationOrderService` and `IEmailNotificationOrderService` to `OrderRequestService` and `IOrderRequestService`.
- Renamed `RegisterEmailNotificationOrder` in `OrderRequestService` to `RegisterNotificationOrder`.
- Added default sender number in the config
- Extended method for setting default value when empty
- Necessary unit tests added
- No manual QA would be done as mentioned in the regarding issue that it is not possible as no external endpoints are available

## Related Issue(s)
- #336

## Verification
  - Except the SonarCloud Code Analysis 
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
+++ b/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
@@ -16,7 +16,7 @@ public class SmsTemplate : INotificationTemplate
     public string SenderNumber { get; internal set; } = string.Empty;
 
     /// <summary>
-    /// Gets the body of emails created by the template    
+    /// Gets the body of SMSs created by the template    
     /// </summary>
     public string Body { get; internal set; } = string.Empty;
 
--- a/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
+++ b/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
@@ -13,7 +13,7 @@ public class SmsTemplate : INotificationTemplate
     /// <summary>
     /// Gets the number from which the SMS is created by the template    
     /// </summary>
-    public string SenderHandle { get; internal set; } = string.Empty;
+    public string SenderNumber { get; internal set; } = string.Empty;
 
     /// <summary>
     /// Gets the body of SMSs created by the template    
@@ -23,9 +23,9 @@ public class SmsTemplate : INotificationTemplate
     /// <summary>
     /// Initializes a new instance of the <see cref="SmsTemplate"/> class.
     /// </summary>
-    public SmsTemplate(string? senderHandle, string body)
+    public SmsTemplate(string? senderNumber, string body)
     {
-        SenderHandle = senderHandle ?? string.Empty;
+        SenderNumber = senderNumber ?? string.Empty;
         Body = body;
         Type = NotificationTemplateType.Sms;
     }
--- a/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
+++ b/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
@@ -3,7 +3,7 @@ using Altinn.Notifications.Core.Enums;
 namespace Altinn.Notifications.Core.Models.NotificationTemplate;
 
 /// <summary>
-/// Template for a SMS notification
+/// Template for an SMS notification
 /// </summary>
 public class SmsTemplate : INotificationTemplate
 {
--- a/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
+++ b/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
@@ -13,7 +13,7 @@ public class SmsTemplate : INotificationTemplate
     /// <summary>
     /// Gets the number from which the SMS is created by the template    
     /// </summary>
-    public string SenderNumber { get; internal set; } = string.Empty;
+    public string SenderHandle { get; internal set; } = string.Empty;
 
     /// <summary>
     /// Gets the body of SMSs created by the template    
@@ -23,9 +23,9 @@ public class SmsTemplate : INotificationTemplate
     /// <summary>
     /// Initializes a new instance of the <see cref="SmsTemplate"/> class.
     /// </summary>
-    public SmsTemplate(string? senderNumber, string body)
+    public SmsTemplate(string? senderHandle, string body)
     {
-        SenderNumber = senderNumber ?? string.Empty;
+        SenderHandle = senderHandle ?? string.Empty;
         Body = body;
         Type = NotificationTemplateType.Sms;
     }
--- a/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
+++ b/src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
@@ -1,0 +1,39 @@
+using Altinn.Notifications.Core.Enums;
+
+namespace Altinn.Notifications.Core.Models.NotificationTemplate;
+
+/// <summary>
+/// Template for a SMS notification
+/// </summary>
+public class SmsTemplate : INotificationTemplate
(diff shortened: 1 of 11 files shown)


src/Altinn.Notifications.Core/Services/OrderRequestService.cs
```diff
@@ -10,33 +10,35 @@
...
 /// </summary>
-public class EmailNotificationOrderService : IEmailNotificationOrderService
+public class OrderRequestService : IOrderRequestService
 {
     private readonly IOrderRepository _repository;
     private readonly IGuidService _guid;
     private readonly IDateTimeService _dateTime;
-    private readonly string _defaultFromAddress;
+    private readonly string _defaultEmailFromAddress;
+    private readonly string _defaultSenderNumber;
```
acn-sbuad, line 21: How about this? 
```suggestion
    private readonly string _defaultSmsSender;
```

src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
```diff
@@ -0,0 +1,39 @@
+using Altinn.Notifications.Core.Enums;
+
+namespace Altinn.Notifications.Core.Models.NotificationTemplate;
+
+/// <summary>
+/// Template for a SMS notification
```
acn-sbuad, line 6:
```suggestion
/// Template for an SMS notification
```

src/Altinn.Notifications.Core/Services/OrderRequestService.cs
```diff
@@ -10,33 +10,35 @@
...
 
     /// <inheritdoc/>
-    public async Task<(NotificationOrder? Order, ServiceError? Error)> RegisterEmailNotificationOrder(NotificationOrderRequest orderRequest)
+    public async Task<(NotificationOrder? Order, ServiceError? Error)> RegisterNotificationOrder(NotificationOrderRequest orderRequest)
     {
         Guid orderId = _guid.NewGuid();
         DateTime created = _dateTime.UtcNow();
 
-        var templates = SetFromAddressIfNotDefined(orderRequest.Templates);
+        var templates = SetFromAddressOrSenderNumberIfNotDefined(orderRequest.Templates);
```
acn-sbuad, line 41: To simplify the method name could we just call it: SetSenderIfNotDefined? This would encompass both from adress and sms sender.

src/Altinn.Notifications.Core/Services/Interfaces/IOrderRequestService.cs
```diff
@@ -6,12 +6,12 @@ namespace Altinn.Notifications.Core.Services.Interfaces;
 /// <summary>
 /// Interface for the email notification order service
```
acn-sbuad, line 7: Comment should be updated to reflect that this does not just support email orders

src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
```diff
@@ -0,0 +1,39 @@
...
+/// </summary>
+public class SmsTemplate : INotificationTemplate
+{
+    /// <inheritdoc/>
+    public NotificationTemplateType Type { get; internal set; }
+
+    /// <summary>
+    /// Gets the number from which the SMS is created by the template    
+    /// </summary>
+    public string SenderNumber { get; internal set; } = string.Empty;
```
acn-sbuad, line 16: Same might apply here? Do we want to use the word number or could we use something else ?

codemascot: Could `SenderHandle` be an option? 🤔

codemascot: Going with `SenderHandle` as Terje approved in the daily standup. So marking this as resolved.

tba76: SenderNumber describes the property better than SenderHandle in my opinion. 
The documentation should describe that this could be number or string.

acn-sbuad: Regardless of what we end up with I think the same term should be used in NotificationOrderConfig and the SMS template as they refer to the same concept

src/Altinn.Notifications.Core/Services/OrderRequestService.cs
```diff
@@ -53,11 +55,16 @@ public async Task<(NotificationOrder? Order, ServiceError? Error)> RegisterEmail
...
         foreach (var template in templates.OfType<EmailTemplate>().Where(template => string.IsNullOrEmpty(template.FromAddress)))
         {
-            template.FromAddress = _defaultFromAddress;
+            template.FromAddress = _defaultEmailFromAddress;
+        }
+
+        foreach (var template in templates.OfType<SmsTemplate>().Where(template => string.IsNullOrEmpty(template.SenderHandle)))
+        {
+            template.SenderHandle = _defaultSmsSender;
         }
```
tba76, line 69: I think this could be refactored so the list will be looped only once and still keep the readability.

codemascot: 1. Firstly, I'm a bit curious how that can be achieved 🤔 Are you thinking about using conditionals? An example would be really helpful. 
2. Secondly, how running a single loop is better in this context? 🤔 So far, in this context, we are querying the templates to get only those which has empty property, and then these goes in the loop. Therefore, isn't running one loop will still go through the same number of elements as running two loops? 🤔

tba76: I was a bit fast when I wrote this comment. Linq will of course filter out the temlates. 
So it all will come to readability, not performance.
I thought about using conditions but I tried to refactor and the result was readable but more code. So keep it as it is.

Example jusf for "fun":
foreach (var template in templates)
{
    switch (template)
    {
        case EmailTemplate emailTemplate when string.IsNullOrEmpty(emailTemplate.FromAddress):
            emailTemplate.FromAddress = _defaultEmailFromAddress;
            break;
        case SmsTemplate smsTemplate when string.IsNullOrEmpty(smsTemplate.SenderHandle):
            smsTemplate.SenderHandle = _defaultSmsSender;
            break;
    }
}

SandGrainOne: In this case both loops goes through all elements in the list at least once. LINQ doesn't remember which elements that were taken care of in the previous iteration and would need to start over. 

Logically this code would generate four loops, but because of deferred execution and iterator support in LINQ and in foreach there is probably just two loops.

src/Altinn.Notifications.Core/Models/NotificationTemplate/SmsTemplate.cs
codemascot: - [Renamed SenderHandle to SenderNumber again as agreed](https://github.com/Altinn/altinn-notifications/pull/380/commits/c0c20dba91e2f1de7bf1729dc4b99b63202b7f13)
- [Renamed NotificationOrderConfig default proporty for SMS as agreed](https://github.com/Altinn/altinn-notifications/pull/380/commits/1cfa0a7f4a977e0a50c3a98def309a3930e92de3)

codemascot:
Had a verbal review from @tba76 to [rename _orderService to _orderRequestService](https://github.com/Altinn/altinn-notifications/pull/380/commits/b843f5f9fc14fd8ef1b11df048dde428799adad2).

tba76 (approved):
🍺

SandGrainOne:
Please try to improve the test coverage a litt bit.

Issue #328, opened 2023-12-06
API logic for placing a SMS notifications order


SandGrainOne:
## Description
Create a new endpoint for placing an order for an SMS notification.

Controller name SMSNotificationOrderController
Endpoint: POST notfications/api/v1/orders/sms

## Considerations
- [x] Assume norwegian number if missing country code? Copy validation rules from Altinn 2
- [x] Are there any numbers we should deny? Copy rules from Altinn 2?
- [x] Text length validation, max length? Splitting of messages on our side?

Fått fra Altinn 2:
``` xml
<add key="KofuviValidCountryCode" value="(^\+([0-9]{1,3}))" />
<add key="KofuviValidMobileNumber" value="(^[0-9]+$)" />
Samt to lister over lovlige landkoder og telefonnummer for hvert land bare Norge har liste over lovlige serier:
[...]

PR #384, opened 2024-01-22
#328 Implementation of API logic for placing a SMS notifications order

codemascot:
Implementation of API logic for placing a SMS notifications order

## Description
- Added `PhoneNumber` property to `RecipientExt` class
- Added `SmsNotificationOrderRequestExt` to represent an SMS notification order request externally for customers
- SMS added as an address type to `AddressType` enum.
- An address point for SMS added
- SMS notification orders controller added
- Attribute added for SMS address point
- Tests for SMS request validator added

## Related Issue(s)
- #328 

## Verification
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)


src/Altinn.Notifications/Controllers/SmsNotificationOrdersController.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <returns>The id of the registered notification order</returns>
+    [HttpPost]
+    [Consumes("application/json")]
+    [Produces("application/json")]
+    [SwaggerResponse(202, "The notification order was accepted", typeof(OrderIdExt))]
+    [SwaggerResponse(400, "The notification order is invalid", typeof(ValidationProblemDetails))]
+    [SwaggerResponseHeader(202, "Location", "string", "Link to access the newly created notification order.")]
+    public async Task<ActionResult<OrderIdExt>> Post(SmsNotificationOrderRequestExt smsNotificationOrderRequest)
+    {
+        var validationResult = _validator.Validate(smsNotificationOrderRequest);
```
SandGrainOne, line 61: One rule we have is to keep declarations explicit when it comes to types.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,55 @@
...
+            .WithMessage("A valid mobile number must be provided for all recipients.");
+
+        RuleFor(order => order.RequestedSendTime)
+            .Must(sendTime => sendTime.Kind != DateTimeKind.Unspecified)
+            .WithMessage("The requested send time value must have specified a time zone.")
+            .Must(sendTime => sendTime >= DateTime.UtcNow.AddMinutes(-5))
+            .WithMessage("Send time must be in the future. Leave blank to send immediately.");
+
+        RuleFor(order => order.Body).NotEmpty();
+        RuleFor(order => order.SenderNumber).NotEmpty();
```
acn-sbuad, line 32: Sender number is not a requirement, they can use default sender `Altinn`
```suggestion
```

src/Altinn.Notifications/Controllers/SmsNotificationOrdersController.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        string selfLink = registeredOrder!.GetSelfLink();
+        return Accepted(selfLink, new OrderIdExt(registeredOrder!.Id));
+    }
+
+    /// <summary>
+    /// Maps a <see cref="SmsNotificationOrderRequestExt"/> to a <see cref="NotificationOrderRequest"/>
+    /// </summary>
+    public static NotificationOrderRequest MapToOrderRequest(SmsNotificationOrderRequestExt extRequest, string creator)
```
acn-sbuad, line 91: I would like to see this logic in a separate mapper class or as a part of one of the classes we're mapping between. Makes more sense to call `NotificationOrderRequest` or `SmsNotificationOrderRequestExt` than `SmsNotificationsOrdersControllers` when needing to map between the two models outside of the controller.

test/Altinn.Notifications.Tests/Notifications/TestingMappers/OrderMapperTests.cs
```diff
@@ -72,7 +72,7 @@ public void MapToNotificationOrderExt_AreEquivalent()
     }
 
     [Fact]
-    public void MapToRecipientExt_AllPropertiesPresent_AreEquivalent()
+    public void MapToRecipientExt_ForEmailAllPropertiesPresent_AreEquivalent()
```
acn-sbuad, line 75: Could we keep the original name and add the sms adress point to the address list instead of duplicating the test?

src/Altinn.Notifications/Models/SmsNotificationOrderRequestExt.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Altinn.Notifications.Models;
+
+/// <summary>
+/// Class representing an SMS notiication order request
+/// </summary>
+/// <remarks>
+/// External representaion to be used in the API.
```
acn-sbuad, line 10:
```suggestion
/// External representation to be used in the API.
```
Please fix the typo in the email class as well

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+        RuleFor(order => order.Body).NotEmpty();
+    }
+
+    /// <summary>
+    /// Validated as mobile number based on the Altinn 2 regex
+    /// </summary>
+    /// <param name="mobileNumber">The string to validate as an mobile number</param>
+    /// <returns>A boolean indicating that the mobile number is valid or not</returns>
+    internal static bool IsValidMobileNumber(string? mobileNumber)
+    {
```
acn-sbuad, line 40: given a Norwegian number +47 or 0047 we want to further validate that the number starts with 9 or 4. We might also want to confirm that it is 8 digits excluding the country code, not sure if they have this last validation in Altinn 2.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+        if (string.IsNullOrEmpty(mobileNumber))
+        {
+            return false;
+        }
+
+        string mobileNumberRegexPattern = @"^(([0-9]{5})|([0-9]{8})|(00[0-9]{3,})|(\+[0-9]{3,}))$";
```
acn-sbuad, line 46: Did we also want to match against the list of valid country codes?

test/Altinn.Notifications.Tests/Notifications/TestingValidators/SmsNotificationOrderRequestValidatorTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            Recipients = new List<RecipientExt>() { new RecipientExt() { MobileNumber = "+4740000001" } },
+            Body = "This is an SMS body"
+        };
+
+        var actual = _validator.Validate(order);
+        Assert.True(actual.IsValid);
+    }
+
+    [Fact]
+    public void Validate_InvalidSmsFormatProvided_ReturnsFalse()
```
acn-sbuad, line 38: If we are referencing the mobile number being invalid, maybe we can use number instead of SMS in the test name ?

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
codemascot: Good point! 🤔 So far I remember, Henning pointed out that the input phone number will always come validated as we are not the public facing API 🤔 Therefore we don't need that much strict validation. 🤔 So I used the Altinn 2 regex only to save time. 🤔 Please correct me if I'm wrong.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
codemascot: Same as the other one! 🤔 https://github.com/Altinn/altinn-notifications/pull/384#discussion_r1466201404

test/Altinn.Notifications.Tests/Notifications/TestingMappers/OrderMapperTests.cs
codemascot: Well, Keeping separate test makes it easier to point out where the failure is happening in the case of failure. Of course we have error message, but when the codebase is getting bigger and complicated it gets hard to debug with error message and in that case separate tests comes quite handy. 🤔 

But I'm also curious about what we can achieve by not duplicating the test, or making one test instead of two? 🤔

src/Altinn.Notifications/Controllers/SmsNotificationOrdersController.cs
codemascot: My actual intention is to make this `MapToOrderRequest` a `private` non-static method of the same class. Why? because it's curated specifically for this class. And it isn't used anywhere else. So why should we overload this method unnecessary! 🤔 The same goes for the `MapToOrderRequest` overloaded method for `EmailNotificationOrderRequestExt` which we are using in a class and it's more suitable for being a helper method than overloaded one. 🤔 So for the sake of maintainability wouldn't it be better to use them as helper method? 🤔  

https://github.com/Altinn/altinn-notifications/blob/cfa975c301bc8003eeb8058e008bb8b7303ff68e/src/Altinn.Notifications/Mappers/OrderMapper.cs#L19

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
acn-sbuad: This is a public facing API, so we should have full validation.

test/Altinn.Notifications.Tests/Notifications/TestingMappers/OrderMapperTests.cs
acn-sbuad: We will at some point support sending through both communication channels. So we also need to confirm that the combination of email and sms works. If the combination works we can trust that they will work separately as well. So I would say 1 test is sufficient if not we should have 3 testing email, sms and both.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
SandGrainOne: This first version will take phon number as input from the service owner requesting the notification. We will in the future add the option to have Notifications do a phone number lookup. Might have been me that managed to create some confusion here, sorry.

codemascot: So what to do now? Should I spend more time on implementing those validation? Or keep it as it is? 🤔 

@SandGrainOne @acn-sbuad

SandGrainOne: Implement validation as good as possible. Use the Altinn 2 logic as inspiration.

src/Altinn.Notifications/Controllers/SmsNotificationOrdersController.cs
acn-sbuad: Making the method private will not allow us to test the mapping isolated, which I think would be a good thing if we could. Grouping mapping logic together in a class will allow us to only navigate to a single file to find all mappings related to the order request. My point for view is that we can benefit by sharing the logic for mapping between the more complex shared types in the order request e.g. recipient. We will have more complex orders in the future that will require addressing both sms and email properties in the same mapping.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
codemascot: This is implemented!

src/Altinn.Notifications/Mappers/OrderMapper.cs
```diff
@@ -16,7 +16,7 @@ public static class OrderMapper
     /// <summary>
     /// Maps a <see cref="EmailNotificationOrderRequestExt"/> to a <see cref="NotificationOrderRequest"/>
     /// </summary>
-    public static NotificationOrderRequest MapToOrderRequest(this EmailNotificationOrderRequestExt extRequest, string creator)
+    public static NotificationOrderRequest ForEmailMapToOrderRequest(this EmailNotificationOrderRequestExt extRequest, string creator)
```
acn-sbuad, line 19: A bit unsure why we need to rename this ForEmail, the type it maps from/to is given based on the type of the first input parameter.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// <param name="mobileNumber">The string to validate as an mobile number</param>
+    /// <returns>A boolean indicating that the mobile number is valid or not</returns>
+    internal static bool IsValidMobileNumber(string? mobileNumber)
+    {
+        if (string.IsNullOrEmpty(mobileNumber))
+        {
+            return false;
+        }
+
+        string mobileNumberRegexPattern = @"^(?:(\+47[49]\d{7})|(0047[49]\d{7})|(?!((\+47[0-9]*)|(0047[0-9]*)))(([0-9]{5})|([0-9]{8})|(00[0-9]{3,})|(\+[0-9]{3,})))$";
```
acn-sbuad, line 47: LinkMobility requires phone numbers to be at least 9 digits. So I guess we can have this validation reflect this as well by removing the allowance for 5 digit numbers

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+                mobileNumber = "+" + mobileNumber.Remove(0, 2);
+            }
+            else
+            {
+                mobileNumber = "+47" + mobileNumber;
+            }
+        }
+
+        PhoneNumber phoneNumber = phoneNumberUtil.Parse(mobileNumber, null);
+        return phoneNumberUtil.IsValidNumber(phoneNumber);
```
acn-sbuad, line 73: would the phoneNumberUtil catch most cases that our regex also catches? Wondering if we could call that first, and then only validate the norwegian number starting with 4 or 9 in the later check? But that depends on how thorough the Util class is.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+            return false;
+        }
+
+        PhoneNumberUtil phoneNumberUtil = PhoneNumberUtil.GetInstance();
+
+        if (!mobileNumber.StartsWith('+'))
+        {
+            if (mobileNumber.StartsWith("00"))
+            {
+                mobileNumber = "+" + mobileNumber.Remove(0, 2);
```
acn-sbuad, line 64: a bit unsure about the maipulation of the phone number during validation. 
Because now it is modifying AND validating rather than just validating. Would it be expensive to move the manipulation of the data to the mapping method instead?

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
acn-sbuad: Unresolving, think this is still relevant. Will this cover everything the regex covers ?

tba76: After we decided to require + or 00 in the api I think the regex validation is unnecessary. I tested very little, but the library should validate against norwegian rules. I tried with a number starting on 5 and that failed. 
Test the library enough to see that it covers what we validates in the regex, then we could drop the regex.

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SmsNotificationOrderRequestValidator"/> class.
+    /// </summary>
+    public SmsNotificationOrderRequestValidator()
+    {
+        RuleFor(order => order.Recipients)
+            .NotEmpty()
+            .WithMessage("One or more recipient is required.")
+            .Must(recipients => recipients.TrueForAll(a => IsValidMobileNumber(a.MobileNumber)))
+            .WithMessage("A valid mobile number must be provided for all recipients.");
```
acn-sbuad, line 24:
```suggestion
            .WithMessage("A valid mobile number istarting with country code must be provided for all recipients.");
```

Will result in a test or two needding to be updated as well

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
codemascot: I also think the RegEx is unnecessary now as the library is pretty smart! So just using the library can achieve what we are doing now I guess. 🤔

test/Altinn.Notifications.Tests/Notifications/TestingValidators/SmsNotificationOrderRequestValidatorTests.cs
```diff
@@ -0,0 +1,167 @@
...
+    [InlineData("90000000", false)]
+    [InlineData("+4790000000", true)]
+    [InlineData("+4750000004", false)]
+    [InlineData("+47900000001", false)]
+    [InlineData("+14790000000", false)]
+    [InlineData("004790000002", true)]
+    [InlineData("", false)]
+    [InlineData("111100000", false)]
+    [InlineData("dasdsadSASA", false)]
+    public void IsValidMobileNumber(string email, bool expectedResult)
```
acn-sbuad, line 162:
```suggestion
    public void IsValidMobileNumber(string mobileNumber, bool expectedResult)
```

src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs
codemascot: So far all the checks has been passed! So we will be relying on the library from now on and dropping the RegEx. :)

acn-sbuad (approved):
🎈

PR #408, opened 2024-02-02
Using Result instead of tuple where applicable

acn-sbuad:
## Description

Removed use of tuples to return success or error response. 

## Verification
- [x] All tests run green

### DIFF
--- a/src/Altinn.Notifications.Core/Services/OrderRequestService.cs
+++ b/src/Altinn.Notifications.Core/Services/OrderRequestService.cs
@@ -1,9 +1,9 @@
 ﻿using Altinn.Notifications.Core.Configuration;
-using Altinn.Notifications.Core.Models;
 using Altinn.Notifications.Core.Models.NotificationTemplate;
 using Altinn.Notifications.Core.Models.Orders;
 using Altinn.Notifications.Core.Persistence;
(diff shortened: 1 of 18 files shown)


src/Altinn.Notifications.Core/Services/OrderRequestService.cs
```diff
@@ -33,7 +33,7 @@ public OrderRequestService(IOrderRepository repository, IGuidService guid, IDate
     }
 
     /// <inheritdoc/>
-    public async Task<(NotificationOrder? Order, ServiceError? Error)> RegisterNotificationOrder(NotificationOrderRequest orderRequest)
+    public async Task<Result<NotificationOrder, ServiceError>> RegisterNotificationOrder(NotificationOrderRequest orderRequest)
```
tba76, line 36: Do we need ServiceError here, it will never be returned?

acn-sbuad: Saving order to db could fail, but we don't catch the exception for now. So either we catch exceptions from the repository layer and return service error or drop the service error.

tba76: Drop is good :)

tba76 (approved):
🍺

Issue #415, opened 2024-02-06
Implement core logic for sms notification summary


acn-sbuad:
### Description

Sms notification summary should be made available for end users through the API and depict the status of the individual notifications generated by an order

### Tasks

- [x] Implementation 
- [x] QA
- [x] Regression tests

### Acceptance Criterias

- [x] Interfaces are defined for persistence to implement
- [x] Models related to sms summary are defined
- [x] Interface and implementation of services are made available to API layer

Issue #414, opened 2024-02-06
Implement repository logic for sms notification summary


acn-sbuad:
### Description

Sms notification summary should be made available for end users through the API and depict the status of the individual notifications generated by an order

### Tasks
- [x] Implementation 
- [x] QA
- [x] Regression tests

### Acceptance Criteria

- [x] Repository method is made available to use by the Core layer

PR #416, opened 2024-02-06
Core + persistence logic for sms notification summaries #GPCActive

acn-sbuad:
## Related Issue(s)
- #414
- #415

## Verification
- [x] All tests run green


src/Altinn.Notifications.Persistence/Repository/NotificationSummaryRepository.cs
```diff
@@ -36,12 +37,70 @@ public NotificationSummaryRepository(NpgsqlDataSource dataSource, TelemetryClien
...
+
+        if (!matchFound)
+        {
+            return null;
+        }
+
+        return new SmsNotificationSummary(orderId)
+        {
+            SendersReference = sendersReference,
+            Notifications = notificationList.Cast<SmsNotificationWithResult>().ToList()
```
SandGrainOne, line 93: Have you tested this without the cast? Looks like the return from GetSummary is of the correct because of clever use of Generics. 

I also tried to add a parameter to `GetSummary` of type `INotificationSummary<T>`. I believe that could work if it wasn't for the fact that the `INotificationSummary.SendersReference` is read only. Could we have used a base class instead of an interface?

You're doing to cool stuff here.

codemascot (approved):
🎉

SandGrainOne (approved):
Just the one comment. Looking good.

src/Altinn.Notifications.Core/Services/SmsNotificationSummaryService.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// <summary>
+    /// Implementation of <see cref="ISmsNotificationSummaryService"/>
+    /// </summary>
+    public class SmsNotificationSummaryService : ISmsNotificationSummaryService
+    {
+        private readonly INotificationSummaryRepository _summaryRepository;
+        private readonly static Dictionary<SmsNotificationResultType, string> _smsResultDescriptions = new()
+        {
+            { SmsNotificationResultType.New, "The sms has been created, but has not been picked up for processing yet." },
```
codemascot, line 17: Just a overall thing. Should we use sms or SMS as SMS is an abbreviation? 🤔

src/Altinn.Notifications.Persistence/Repository/NotificationSummaryRepository.cs
acn-sbuad: I tried to convert it to a base class instead of interface, but end ut with a cast exception when trying to return the final summary class explicitly. Merging feature for now, and will try a bit more on further simplifying the repository logic later today.

PR #420, opened 2024-02-08
Patched Altinn.Common.AccessToken dependency

acn-sbuad:
## Description

Patched library with bug fix for excessive logging. 

## Related Issue(s)
- N/A

### DIFF
--- a/src/Altinn.Notifications/Altinn.Notifications.csproj
+++ b/src/Altinn.Notifications/Altinn.Notifications.csproj
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Altinn.Common.AccessToken" Version="3.0.3" />
+    <PackageReference Include="Altinn.Common.AccessToken" Version="4.1.0-preview" />
     <PackageReference Include="Altinn.Common.PEP" Version="1.3.0" />
     <PackageReference Include="FluentValidation" Version="11.9.0" />
     <PackageReference Include="JWTCookieAuthentication" Version="4.0.1" />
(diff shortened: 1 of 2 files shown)


src/Altinn.Notifications/Altinn.Notifications.csproj
```diff
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Altinn.Common.AccessToken" Version="3.0.3" />
+    <PackageReference Include="Altinn.Common.AccessToken" Version="4.1.0" />
```
SandGrainOne, line 13: Now that I understand a bit better what it is that is breaking Authorization I believe this should work. Notification doesn't have ABAC based Authorization as far as I can remember.

For application with ABAC based Authorization I hope updating PEP to 1.4.0-alpha3 will fix the issues we've been having. We can try this in Events if you want.

test/Altinn.Notifications.Tests/Altinn.Notifications.Tests.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Altinn.Common.AccessToken" Version="3.0.3" />
+    <PackageReference Include="Altinn.Common.AccessToken" Version="4.1.0" />
```
SandGrainOne, line 10: This reference should be unnecessary.

SandGrainOne (approved):

src/Altinn.Notifications/Altinn.Notifications.csproj
acn-sbuad: will test in events.

test/Altinn.Notifications.Tests/Altinn.Notifications.Tests.csproj
acn-sbuad: A package downgrade is introduced if we don't have this explicitly. Will keep for now.

Issue #361, opened 2024-01-03
Implement support for LinkMobility status codes 


tba76:
### Description

Workflow: Update status for an SMS notifications will need modifications to the persistence layer of the notification app. 

PSWinCom Gateway Delivery Report Statuses:
DELIVRD: Message was successfully delivered to destination.
EXPIRED: Message validity period has expired.
DELETED: Message has been deleted.
UNDELIV: The SMS was undeliverable (not a valid number or no available route to destination).
UNKNOWN: No information of delivery status available.
REJECTD: Message was rejected.
FAILED: The SMS failed to be delivered because no operator accepted the message or due to internal Gateway error.
NULL: No delivery report received from operator. Unknown delivery status.
BARRED: The receiver number is barred/blocked/not in use. Do not retry message, and remove number from any subscriber list.

### Consideration
- [x] Do we need to log the total number of SMS that the notification generated?
  - Yes, extracted in seperate issue #440

### Tasks
- [x] Extend _SmsNotificationResultType_ with new result types
- [x] Extend _smsnotificationresulttype_ enum in db with new result types
- [x] Add descriptions for new types in _NotificationSummaryService_
- [x] Implement unit tests
- [x] Implement integration tests
- [x] QA
- [x] Manual testing (if API endpoint is available)
- [x] Documentation in developer guides on what each status code means

### Acceptance Criterias
- [x]  Regression tests

acn-sbuad:
Verified in TT

PR #444, opened 2024-02-19
option to update sms send status with only gateway ref

acn-sbuad:
## Description

When delivery reports are returned for sms, the notificationId is not available, hence it should be possible to update the database with only the gateway reference

## Related Issue(s)
- #361

## Verification
- [x] All tests run green


tba76 (approved):
🍺

PR #474, opened 2024-03-21
Better validation of notification orders

martinothamar:
## Description
Implements validation of more order fields based on constraints from downstream providers. I came across this since I was wondering what the input constraints was when developing the SMS client in apps

## Related Issue(s)
N/A

## Verification
- [ ] All tests run green

## Documentation


martinothamar:
@acn-sbuad since this would be a breaking change I'm not sure if this is something you want. But I think a change like this is good because
* Users get bad request errors early (as opposed to only noticing when reading the status, I'm assuming errors due to service limits would appear somewhere in the status/delivery reports).
* Unbounded input is not a good idea in terms of security, upstream applications could affect service availability and performance (either due to bugs or security issues) I think

If we do something like this, we should document the limits as well

src/Altinn.Notifications/Validators/EmailNotificationOrderRequestValidator.cs
```diff
@@ -17,8 +17,8 @@ public class EmailNotificationOrderRequestValidator : AbstractValidator<EmailNot
     public EmailNotificationOrderRequestValidator()
     {
         RuleFor(order => order.Recipients)
-            .NotEmpty()
-            .WithMessage("One or more recipient is required.")
+            .Must(r => r?.Count is >= 1 and <= 50) // Azure Communication Services has max 10 recipients
```
acn-sbuad, line 20: Number of recipients for an order doesn't directly translate to how many recipients are set in the to field of an email. One email is sent per recipient.

martinothamar: Ahh, makes sense! Should there be still some upper bound? Like should [Array.MaxLength (2,147,483,591)](https://learn.microsoft.com/en-us/dotnet/api/system.array.maxlength?view=net-8.0) number of recipients be allowed?

acn-sbuad: I would like to keep it open for now. We are still working on analyzing how batch orders should be handled and will most likely have a separate endpoint for posting extra large orders (think information about tax forms being available), once this has been specified we will be able to set a reasonable limit for this endpoint. Current use cases, e.g. notifications from an app triggered by the user or by instantiation will not generate any significant number of recipients is our current hypothesis.

PR #475, opened 2024-03-21
Ensure notification orders are placed within transaction

martinothamar:
## Description
I noticed that orders are not included in the transaction when executed. [When `CreateCommand` is invoked on `NpgsqlDataSource`, it is creating a new connection](https://github.com/npgsql/npgsql/blob/9e4ff8ce78c586ec19c1ae7ef1ccff29d08a1c39/src/Npgsql/NpgsqlDataSource.cs#L202-L203)

I also awaited each secondary insert as well, as the connection itself is not threadsafe

## Related Issue(s)
N/A

## Verification
- [ ] All tests run green

## Documentation


acn-sbuad (approved):
🎈

PR #488, opened 2024-04-11
#439 only persist national identity number on notification

acn-sbuad:
## Description

- Fixed bug where national identity number is persisted in db on the generated notification
- Changed implementation to ensure that intial lookup of contact details & reservation is not persisted to the notification order in db. 

### Consideration: 

Merging this PR will result in reservation status being made available for TE in order request response, and through orders/{id}/notifications/sms endpoint. 

  - The orders/{id} returns original order without reserve status or contact details. Should this endpoint rather check if notifications have been generated and merge the original order with the recipient data we used when actually sending the notification? How would this look for an org, where a single order recipient results in multiple notification recipients? 

## Related Issue(s)
- #439

### DIFF
--- a/src/Altinn.Notifications.Core/Models/Recipient.cs
+++ b/src/Altinn.Notifications.Core/Models/Recipient.cs
@@ -41,18 +41,18 @@ public class Recipient
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


tba76 (approved):
🍺

Issue #472, opened 2024-03-21
Add access based recipient filtering of persons with personal contact information associated with an organsation


SandGrainOne:
### Description

Create a service responsible for authorization. Add logic to check that the person to receive the notification are allowed to represent the organization in the context of the given resource. If no resource or app has been given as context of the notification, then personal contact information should not be considered. All authorization requires a resource.

Perform a normal authorization request similar to the requests created by Storage when a user wants to access a resource. 

### Tasks

- [x] Create a new service class to handle Authorization 
- [x] Create a method that can determine if a set of users are allowed to access given resource for given subject.
- [x] Unit tests
- [x] QA
- [x] Manual testing
- [x] Update [architecture documentation](https://docs.altinn.studio/notifications/architecture/notifications/) with new dependency

### Acceptance criteria

- [x] Service is available for use by logic in other services.

acn-sbuad:
Satisfied the authorization logic works as expected after testing in AT and locally.

PR #520, opened 2024-05-10
Authorization based filtering of recipients (#472)

SandGrainOne:
## Description
Adding authorization logic to ensure that the recipients of notifications can access the resource they're being notified about.

My own thoughts is that we probably don't need to return a dictionary inside a dictionary. It might be enough with a list as the inner collection. The logic removes all results that are not Permit.

## Related Issue(s)
- #472 

## Verification
- [x] New code builds clean without any errors or warnings
- [x] All tests run green

## Documentation


acn-sbuad:
Placing the logic in the API project makes Core dependent on the Notifications project. I would rather at least the interface be defined in core. The service that will use the authorization is placed in core.

src/Altinn.Notifications/appsettings.json
```diff
@@ -1,8 +1,8 @@
 {
   "PlatformSettings": {
     "ApiProfileEndpoint": "http://localhost:5030/profile/api/v1/",
-    "ApiRegisterEndpoint": "http://localhost:5020/register/api/v1/"
-
+    "ApiRegisterEndpoint": "http://localhost:5020/register/api/v1/",
+    "ApiAuthorizationEndpoint": "http://localhost:5030/authorization/api/v1/"
```
acn-sbuad, line 5: I though authorization was port 5050. Will you double check. Shouldn't be the same as profile

```suggestion
    "ApiAuthorizationEndpoint": "http://localhost:5050/authorization/api/v1/"
```

SandGrainOne:
> Placing the logic in the API project makes Core dependent on the Notifications project. I would rather at least the interface be defined in core. The service that will use the authorization is placed in core.

You're right. I forgot to move the interface. I don't want the service logic in Core because it would make Core dependent of PEP. Another idea is to move the logic to Integrations and rename it to "Client". That would isolate the external dependency to a more appropriate library.

src/Altinn.Notifications/appsettings.json
SandGrainOne: Is the Dockerfile in the authorization project wrong then? How does that work? Is there an override somewhere?
`EXPOSE 5030`

src/Altinn.Notifications/Authorization/IAuthorizationService.cs
```diff
@@ -0,0 +1,17 @@
...
+/// Describes the necessary functions of an authorization service that can perform
+/// notification recipient filtering based on authorization
+/// </summary>
+public interface IAuthorizationService
+{
+    /// <summary>
+    /// Describes a method that can create an authorization request to authorize a set of
+    /// users for access to a resource.
+    /// </summary>
+    /// <param name="orgRightHolders">The list organizations with associated right holders.</param>
```
acn-sbuad, line 13: Think we need to be explicit somewhere in what the two ints are. I cannot find partyId and userId mentioned anywhere  at the moment.. We need a more explicit definition of this, so everyone can understand.

src/Altinn.Notifications/appsettings.json
acn-sbuad: Probably. We set the[ port in the helm chart](https://dev.azure.com/brreg/_git/altinn-studio-ops?path=/deploy/altinn-platform/altinn-authorization/values.yaml&version=GBmaster&line=7&lineEnd=8&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) and there 5050 is used for authorization.

SandGrainOne: Okay, but the helm chart is not involved when I run the application in docker/podman.

src/Altinn.Notifications/Program.cs
```diff
@@ -182,6 +186,11 @@ void ConfigureServices(IServiceCollection services, IConfiguration config)
     AddInputModelValidators(services);
     services.AddCoreServices(config);
 
+    services.Configure<Altinn.Common.PEP.Configuration.PlatformSettings>(config.GetSection("PlatformSettings"));
```
acn-sbuad, line 189: Should this whole section rather be added to an extension method in the integrations project? Following the pattern for AltinnClients and CoreServices ?

src/Altinn.Notifications.Core/Integrations/IAuthorizationService.cs
```diff
@@ -0,0 +1,20 @@
...
+/// Describes the necessary functions of an authorization service that can perform
+/// notification recipient filtering based on authorization
+/// </summary>
+public interface IAuthorizationService
+{
+    /// <summary>
+    /// Describes a method that can create an authorization request to authorize a set of
+    /// users for access to a resource.
+    /// </summary>
+    /// <param name="organizationContactPoints">The list organizations with associated right holders.</param>
```
acn-sbuad, line 15:
```suggestion
    /// <param name="organizationContactPoints">The contact points of an organization including user registered contact points.</param>
```

src/Altinn.Notifications.Core/Integrations/IAuthorizationService.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <param name="resourceId">The id of the resource.</param>
+    /// <returns>A task</returns>
```
acn-sbuad, line 17: Returns a filtered list maybe?

src/Altinn.Notifications.Core/Models/ContactPoints/OrganizationContactPoints.cs
```diff
@@ -19,4 +24,9 @@ public class OrganizationContactPoints
     /// Gets or sets a list of official email addresses
     /// </summary>
     public List<string> EmailList { get; set; } = [];
+
+    /// <summary>
+    /// Gets or sets a list of user registered contanct points associated with the organisation.
```
acn-sbuad, line 29:
```suggestion
    /// Gets or sets a list of user registered contact points associated with the organization.
```

src/Altinn.Notifications.Integrations/Authorization/AuthorizationClient.cs
```diff
@@ -0,0 +1,200 @@
...
+using Altinn.Notifications.Core.Models.ContactPoints;
+using static Altinn.Authorization.ABAC.Constants.XacmlConstants;
+
+namespace Altinn.Notifications.Integrations.Authorization;
+
+/// <summary>
+/// An implementation of <see cref="IAuthorizationService"/> able to check that a potential
+/// recipient of a notification can access the resource that the notification is about.
+/// </summary>
+public class AuthorizationClient : IAuthorizationService
```
acn-sbuad, line 17: what about calling it a Service rather than client. The client is implemented and used within PDP I would say, this is just a service using yet another service.

src/Altinn.Notifications/appsettings.json
acn-sbuad: Issue has been fixed in authz repo, so we can update here as well.

src/Altinn.Notifications/Program.cs
```diff
@@ -6,11 +6,15 @@
 using Altinn.Common.AccessToken;
 using Altinn.Common.AccessToken.Services;
 using Altinn.Common.PEP.Authorization;
+using Altinn.Common.PEP.Clients;
```
acn-sbuad, line 9: Are these three lines obsolete now?

src/Altinn.Notifications.Core/Integrations/IAuthorizationService.cs
```diff
@@ -0,0 +1,20 @@
...
+public interface IAuthorizationService
+{
+    /// <summary>
+    /// Describes a method that can create an authorization request to authorize a set of
+    /// users for access to a resource.
+    /// </summary>
+    /// <param name="organizationContactPoints">The contact points of an organization including user registered contact points.</param>
+    /// <param name="resourceId">The id of the resource.</param>
+    /// <returns>A task</returns>
+    Task<List<OrganizationContactPoints>> AuthorizeUsersForResource(
```
acn-sbuad, line 18: Could it be named AuthorizeContactPointsForResource? Or (a but of a mouthful) AuthorizeUserContactPointsForResource ?

src/Altinn.Notifications.Core/Models/ContactPoints/OrganizationContactPoints.cs
```diff
@@ -19,4 +26,30 @@ public class OrganizationContactPoints
...
+    /// Gets or sets a list of user registered contact points associated with the organization.
+    /// </summary>
+    public List<UserContactPoints> UserContactPoints { get; set; } = [];
+
+    /// <summary>
+    /// Create a new instance with the same values as the existing instance
+    /// </summary>
+    /// <returns>The new instance with copied values.</returns>
+    public OrganizationContactPoints CloneWithoutUsers()
+    {
```
acn-sbuad, line 40: wondering if we could clone without any contactPoints, not just excluding user contact points. As it is intended used the email and mobile number list won't be populated either way. Cannot see any use cases where they might be for the time being either

src/Altinn.Notifications.Integrations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -53,6 +57,19 @@ public static void AddAltinnClients(this IServiceCollection services, IConfigura
         services.AddHttpClient<IRegisterClient, RegisterClient>();
     }
 
+    /// <summary>
+    /// Adds Altinn clients and configurations to DI container.
```
acn-sbuad, line 61: comment needs updating

acn-sbuad:
```
System.InvalidOperationException : Unable to resolve service for type 'Microsoft.AspNetCore.Http.IHttpContextAccessor' while attempting to activate 'Altinn.Common.PEP.Clients.AuthorizationApiClient'.
```

Getting this error when plugging this code into the new logic in contact point service. Did you experience this when testing locally? Wondering if `IHttpContextAccessor` should be added to the AddAuthorizationService function.

src/Altinn.Notifications.Integrations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -53,6 +57,19 @@ public static void AddAltinnClients(this IServiceCollection services, IConfigura
...
+    /// </summary>
+    /// <param name="services">service collection.</param>
+    /// <param name="config">the configuration collection</param>
+    public static void AddAuthorizationService(this IServiceCollection services, IConfiguration config)
+    {
+        services.Configure<Common.PEP.Configuration.PlatformSettings>(config.GetSection("PlatformSettings"));
```
acn-sbuad, line 67: This fixed the error on my end

```suggestion
        services.Configure<Common.PEP.Configuration.PlatformSettings>(config.GetSection("PlatformSettings"));
services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```

SandGrainOne: That is strange. The service is already being added in Program.cs line 206.

test/Altinn.Notifications.Tests/Notifications.Integrations/AuthorizationServiceTests.cs
```diff
@@ -0,0 +1,148 @@
+﻿using System.Collections.Generic;
```
acn-sbuad, line 1: could we move this into an Auuthorization folder to match existing structure for tests?
Can also consider moving the Testdata folder to the Notification.Integrations folder to keep it close to the tests that use them

src/Altinn.Notifications.Core/Models/ContactPoints/OrganizationContactPoints.cs
SandGrainOne: I changed it.

acn-sbuad (approved):
🎈 a few code smells, seem easy to fix

PR #528, opened 2024-05-28
only copy recipient if lookup is required

acn-sbuad:
## Description

Optimizing the code. Currently all recipients are deep copied when completing initial lookup, this is only needed for the users that will be looked up. 

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green


tba76 (approved):
🍺

Issue #510, opened 2024-05-06
Add regression and use case tests that include an app affiliation


SandGrainOne:
## Description
Expand the k6 tests to include verification of address lookup when recipient is an organization. The test should include app affiliation to test the lookup for personal contact information. This should also make it unnecessary to keep the NIN lookup test as both Register and Profile will be included in this new test.

We can use the ttd apps-test app for testing app affiliation. Might need to adjust policy.

## Tasks
- [x] Generate or identify test data for address lookup with organisation number that includes an app affiliation.
 
## Acceptance criteria
- [x] Use-case tests include a verification of address lookup of an organization. 
- [x] Regression tests include a verification of address lookup of an organization.
- [x] Remove the NIN lookup test from the use-case test. (Keep it for the regression test.)

Issue #507, opened 2024-05-06
Regression and use case tests - OrgNo lookup


SandGrainOne:
## Description
Expand the k6 tests to include verification of address lookup when recipient is an organization.

## Tasks
- [x] Generate or identify test data for address lookup with organisation number 
 
## Acceptance criteria
- [ ] Use-case tests include a verification of address lookup of an organization. 
- [x] Regression tests include a verification of address lookup of an organization.

SandGrainOne:
Fulfilled by #510

PR #540, opened 2024-06-06
#510 Implementation of adding regression and use case tests that include an app affiliation

codemascot:
#510 Implementation of adding regression and use case tests that include an app affiliation

## Description
#510 Implementation of adding regression and use case tests that include an app affiliation

## Related Issue(s)
- #510
- #507 

## Verification
- [x] All tests run green

### DIFF
--- a/test/k6/src/tests/orders_org_no.js
+++ b/test/k6/src/tests/orders_org_no.js
@@ -17,7 +17,7 @@
     For use case tests omit environment variable runFullTestSet or set value to false
 */
 
-import { check } from "k6";
+import { check, sleep } from "k6";
 import * as setupToken from "../setup.js";
 import * as notificationsApi from "../api/notifications/notifications.js";
 import * as ordersApi from "../api/notifications/orders.js";
@@ -26,11 +26,10 @@ const emailOrderRequestJson = JSON.parse(
   open("../data/orders/01-email-request.json")
 );
 
-import { generateJUnitXML, reportPath } from "../report.js";
 import { addErrorCount, stopIterationOnFail } from "../errorhandler.js";
 const scopes = "altinn:serviceowner/notifications.create";
-const emailRecipient = __ENV.emailRecipient.toLowerCase();
-const ninRecipient = __ENV.ninRecipient.toLowerCase();
+const orgNoRecipient = __ENV.orgNoRecipient.toLowerCase();
+const resourceId = __ENV.resourceId;
 export const options = {
   thresholds: {
     errors: ["count<1"],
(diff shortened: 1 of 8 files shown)


acn-sbuad:
isn't the same resource used in all environments? Would make more sense to have this as a repo secret rather than an env secret.

acn-sbuad:
Is it a conscious decision to both org lookup and resource dependency to the existing test case rather than adding a new test with a seperate order request allowing us to validate the sending of an email to a known recipient isolated from sending to org and resource?

acn-sbuad:
Would be good if we could document the org numbers in pedia along with the other test data.

codemascot:
> isn't the same resource used in all environments? Would make more sense to have this as a repo secret rather than an env secret.

Yes, you are correct! 👍 This is more fit for being a repo secret than an environment variable. 

> Is it a conscious decision to both org lookup and resource dependency to the existing test case rather than adding a new test with a separate order request allowing us to validate the sending of an email to a known recipient isolated from sending to org and resource?

Yes! It was a  conscious decision! Well, although we thought about writing a different test at first, later we agreed to put it together with the existing one as this is a regression test. However, it is definitely worth a consideration again which approach would be better.

> Would be good if we could document the org numbers in pedia along with the other test data.

Already started working on it. Trying to find a better way to document it in AltinnPedia. 👍

codemascot:
@acn-sbuad Here is the doc update: https://github.com/Altinn/altinnpedia/pull/148

test/k6/src/tests/orders_org_no.js
```diff
@@ -0,0 +1,263 @@
...
+ * 04 - GET notification order with status
+ * 05 - GET email notification summary
+ * 06 - GET email notifications summary again after one minute for verification
+ */
+export default function (data) {
+  try {
+    if (data.runFullTestSet) {
+      var selfLink = TC01_PostEmailNotificationOrderRequest(data);
+      let id = selfLink.split("/").pop();
+      TC02_GetNotificationOrderById(data, selfLink, id);
```
acn-sbuad, line 239: not sure I see the point of duplicating test caces 2-5 from the existing orders_email tests. Recipient being an organization doesn't change any of the logic related to retrieving an order.

SandGrainOne: I also think that we can remove some of the duplicated tests.

codemascot: I do agree! Will remove them today! 👍

codemascot: I do agree that 02-04 we don't need! However we do need the 05 for the use case test, don't we? 🤔

SandGrainOne (approved):

PR #564, opened 2024-06-25
Implemented condition client logic

acn-sbuad:
## Related Issue(s)
- #564

## Verification
- [x] All tests run green

### DIFF
--- a/src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
+++ b/src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
@@ -1,7 +1,7 @@
 ﻿namespace Altinn.Notifications.Core.Models.SendCondition
 {
     /// <summary>
-    /// Condition client error class
+    /// Model describing a condition client error
     /// </summary>
     public class ConditionClientError
     {
--- a/src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
+++ b/src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
@@ -1,0 +1,18 @@
+﻿namespace Altinn.Notifications.Core.Models.SendCondition
+{
+    /// <summary>
+    /// Condition client error class
+    /// </summary>
+    public class ConditionClientError
+    {
+        /// <summary>
+        /// The result code of the condition check request
+        /// </summary>
+        public int? StatusCode { get; set; }
+
+        /// <summary>
+        /// The error message of the condition check request
+        /// </summary>
+        public string? Message { get; set; }
+    }
+}
--- a/src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
+++ b/src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
@@ -48,7 +48,7 @@ namespace Altinn.Notifications.Integrations.SendCondition
                 }
             }
 
-            return new ConditionClientError();
+            return new ConditionClientError { StatusCode = (int)res.StatusCode, Message = responseString };
         }
     }
 }
--- a/src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
+++ b/src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
@@ -1,0 +1,54 @@
+﻿using System.Text.Json;
+
+using Altinn.Notifications.Core.Integrations;
+using Altinn.Notifications.Core.Models.SendCondition;
+using Altinn.Notifications.Core.Shared;
+
+namespace Altinn.Notifications.Integrations.SendCondition
+{
+    /// <summary>
+    /// Implementation of <see cref="IConditionClient"/> using an Altinn Client to send the condition request
+    /// </summary>
+    public class SendConditionClient : IConditionClient
+    {
+        private readonly HttpClient _client;
+
+        /// <summary>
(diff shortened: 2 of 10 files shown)


src/Altinn.Notifications/Controllers/SendConditionController.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Controller for supporting automated testing of send condition processing
+    /// </summary>
+    [Route("notifications/api/v1/tests/sendcondition")]
+    [Consumes("application/json")]
+    [SwaggerTag("Private API")]
+    [ExcludeFromCodeCoverage]
+    public class SendConditionController : Controller
+    {
+        /// <summary>
+        /// Accepts an http post request and responds OK.
```
acn-sbuad, line 20:
```suggestion
        /// Condition endpoint that returns a send condition response.
```
```suggestion
        /// Accepts an http post request and responds OK.
```

src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Model describing a condition client error
+    /// </summary>
+    public class ConditionClientError
+    {
+        /// <summary>
+        /// The result code of the condition check request
+        /// </summary>
+        public int? StatusCode { get; set; }
```
SandGrainOne, line 11: Do we need these two properties to be nullable?

src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            if (res.IsSuccessStatusCode)
+            {
+                try
+                {
+                    SendConditionResponse? conditionResponse = JsonSerializer.Deserialize<SendConditionResponse?>(responseString);
+
+                    if (conditionResponse?.SendNotification != null)
+                    {
+                        return (bool)conditionResponse.SendNotification;
```
SandGrainOne, line 38: I'm getting an error if I try to cast null to a bool. Might need to add a check.

src/Altinn.Notifications.Integrations/SendCondition/SendConditionResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Represents the response indicating whether a notification should be sent.
+    /// </summary>
+    public class SendConditionResponse
+    {
+        /// <summary>
+        /// A boolean indicating whether a notification should be sent or not
+        /// </summary>
+        /// <remarks>
+        /// Nullable to ensure default is not failse if deserialization fails
```
SandGrainOne, line 14:
```suggestion
        /// Nullable to ensure default is not false if deserialization fails
```

src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
acn-sbuad: Isn't the check already there? line 36

src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
acn-sbuad: My thought was that status code doesn't always make sense to add.. if response is 200 ok for instance. Or if we get a timeout and there never is a status code. And a message might not make sense to add if we get a 403 for instance, that's why I made them both nullable for now.

test/Altinn.Notifications.Tests/Notifications.Integrations/SendCondition/SendConditionClientTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            // Act
+            Result<bool, ConditionClientError> result = await _sendConditionClient.CheckSendCondition(new Uri("http://test.com?desiredResponse=invalidbody"));
+
+            // Assert
+            result.Match(
+                sendNotification => throw new Exception("No success value should be returned if json deserialization fails"),
+                actuallError =>
+                {
+                    Assert.False(string.IsNullOrEmpty(actuallError?.Message));
```
SandGrainOne, line 108: Should be unnecessary to check if `actualError` is null. The unit tests of `Result<T>` should check the `Result<T>` implementation and ensure that it never match on `actualError` if it's null.

This is repeated a few times in other tests as well.

src/Altinn.Notifications.Integrations/SendCondition/SendConditionClient.cs
SandGrainOne: Sorry, just me being blind. In that case the cast is unnecessary. 
```suggestion
                        return conditionResponse.SendNotification.Value;
```

SandGrainOne: Looks like the compiler produce the same code in both cases. I still prefer to avoid casting when I can.

src/Altinn.Notifications.Core/Models/SendCondition/ConditionClientError.cs
SandGrainOne: Yeah, okay. I can live with that.

Ideally I don't think Core should care about https status codes, but I'm not going to spend the time on defining an abstraction and mapping.

SandGrainOne (approved):

Issue #549, opened 2024-06-11
Add send condition evaluation to order processing


acn-sbuad:
## Description

#547 convers the epic of supporting send conditions in notifications. A part of this is to evaluate the send condition using the condition during the processing of an order to decide if a notification should be sent or not. 

The scope of this issue is to validate, accept/deny and persist the send condition endpoint along with the notification order.  

### Questions: 
could a dialogue be read and deleted before the requested send time? if so what would the response of the send condition endpoint be?  Strongly linked to dialogue element or to a persistent log? 

### Should condition be evaluated before or after a notification is generated?
To be more specific, if the send condition fails is the notification generated at all ? And if not, does the creator need to be made aware that the send condition failed? Is this then persisted in the order status? 

__If a notification IS NOT generated when condition fails__

ACCEPTED Solution: We could consider adding a new status to the OrderProcessing status currently holds values _Registered_, _Processing_, _Completed_. We could add `SendConditionNotMet`.

Pros: 
- We limit the logic to the orderprocessing never involving the notification creation logic.

Cons: 
- The SendConditionNotMet as a part of the orderProcessingStatus this seems a bit unrelated to the actual processing of the order. 

__If a notification IS generated when condition fails__
Solution: we could have a decorator shared between all order processing to lookup send condition to avoid duplication. Handling the buypass of lookup and going directly to generation of notification would need to be handled somehow. This is a need both with the decorator solution and not. 

Pros: 
- We can add the status to the email and sms notificationResultType as `SendConditionNotMet`. A benefit of this solution is that it is explicit for the creator of the notification that send condition was not met. 

Cons: 
- Recipient lookup. But we could in this case avoid the lookup if we know send condition is not met, and only reference the user with the provided organization or national identity number. 
- Metrics. Currently we count each notification being generated as something that needs to be billed. For SMS we can solve it by setting the count to 0, but for emails each generated notification is counted a billable email. We could tweak the script to not count notifications where send condition has not been met. 

### How are errors during condition lookups handled?
__Fail fast__

Pros: 
- No retry count to persists.
 
Cons: 
- Notifications that should be sent are not sent. 
- If error is fixed we might be required to manually trigger processing again.  

__Retry__

Pros: 
- could be a passing error and a retry would fix the issue. 

Cons: 
- at some point we stop retry and the problem with the endpoint might still not be fixed. Retries will then just prolong the processing of an issue. 

__Manual intervention and responsibilities__
[...]

PR #567, opened 2024-07-02
Added condition check to order processing

acn-sbuad:
## Description

Checking send condition during order processing. 
- Failure in first attempt => exception thrown to trigger retry logic
- Failure in retry => order processed

Added delay to order processing.

## Related Issue(s)
- #549 

## Verification
- [x] All tests run green


test/Altinn.Notifications.Tests/Notifications.Core/TestingServices/OrderProcessingServiceTests.cs
```diff
@@ -175,11 +242,78 @@ public async Task ProcessOrderRetry_SerivceThrowsException_ProcessingStatusIsNot
...
+        // Assert
+        Assert.True(actual);
+    }
+
+    [Theory]
+    [InlineData(true, false)]
+    [InlineData(false, false)]
+    [InlineData(true, true)]
+    [InlineData(false, true)]
+    public async Task IsSendConditionMet_SuccessResultFromClient_ReturnsSameAsClient(bool expectedConditionResult, bool isRetry)
```
SandGrainOne, line 264: I'm not sure if the `isRetry` parameter is making a difference. The setup is such that the result match statement will never hit the error case where the `isRetry` value is used.

acn-sbuad: Thought here was to be explicit on expected outcome for all cases, so we are able to catch if behavior of code changes in the future. I didn't set up any other happy path tests (condition is true) for retry.

Issue #562, opened 2024-06-24
Automated tests


SandGrainOne:

acn-sbuad:
Merged and verified test case runs green in all envs in regression test worklfow

PR #568, opened 2024-07-03
Update threshold to use checks metrics instead of errors

SandGrainOne:
## Description
I've been using this time to learn more about k6 in preparation for issue #562 and this pull request is the result.

- Changing the threshold rule for all tests and removing the custom "error" metric. 
- Simplified the `stopInterationOnFail` method and updated call points.

## Related Issue(s)
- #562

## Verification
- [x] Your code builds clean without any errors or warnings

### DIFF
--- a/test/k6/src/tests/orders_email.js
+++ b/test/k6/src/tests/orders_email.js
@@ -1,54 +1,51 @@
 /*
-    Test script of platform notifications api with org token
+    Test script of platform notifications api with application owner token
     Command:
-    docker-compose run k6 run /src/tests/orders_email.js `
+    podman compose run k6 run /src/tests/orders_email.js `
     -e tokenGeneratorUserName=autotest `
-    -e tokenGeneratorUserPwd=*** `
-    -e mpClientId=*** `
-    -e mpKid=altinn-usecase-events `
-    -e encodedJwk=*** `
-    -e env=*** `
-    -e emailRecipient=*** `
-    -e ninRecipient=*** `
+    -e tokenGeneratorUserPwd={the password to access the token generator}`
+    -e mpClientId={the id of an integration defined in maskinporten} `
+    -e mpKid={the key id of the JSON web key used to sign the maskinporten token request} `
+    -e encodedJwk={the encoded JSON web key used to sign the maskinporten token request} `
+    -e env={environment: at21, at22, at23, at24, tt02, prod}`
+    -e emailRecipient={an email address to add as a notification recipient}`
+    -e ninRecipient={a national identity number of a person to include as a notification recipient}`
     -e runFullTestSet=true
 
     For use case tests omit environment variable runFullTestSet or set value to false
 */
 
 import { check } from "k6";
+import { uuidv4 } from "https://jslib.k6.io/k6-utils/1.4.0/index.js";
+import { stopIterationOnFail } from "../errorhandler.js";
+
 import * as setupToken from "../setup.js";
 import * as notificationsApi from "../api/notifications/notifications.js";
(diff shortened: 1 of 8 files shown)


test/k6/src/tests/orders_sms.js
```diff
@@ -122,11 +116,7 @@ function TC03_GetNotificationOrderBySendersReference(data) {
       r.status === 200,
   });
 
-  addErrorCount(success);
-  if (!success) {
-    // only continue to parse and check content if success response code
-    stopIterationOnFail(success);
-  }
+  stopIterationOnFail("GET notification order by senders reference request failed", success);
```
acn-sbuad, line 119: not sure if we need to stop the full run if this request fails? Summaries etc. might still run green ?

test/k6/src/tests/orders_sms.js
```diff
@@ -196,31 +177,14 @@ function TC05_GetSmsNotificationSummary(data, orderId) {
...
-      var selfLink = TC01_PostSmsNotificationOrderRequest(data);
-      let id = selfLink.split("/").pop();
-      TC05_GetSmsNotificationSummary(data, id);
-    }
-  } catch (error) {
-    addErrorCount(false);
-    throw error;
-  }
-}
+  var selfLink = TC01_PostSmsNotificationOrderRequest(data);
```
acn-sbuad, line 180: think I prefer the old notation. If we get to a point where a TC 6 should be a part of the full test set but not the regression tests what do we do? 

It is duplication of code, but more explicit the order of the tests in each case?

test/k6/src/tests/orders_sms.js
acn-sbuad: Would it make sense to only have the test that produces the test data stop the full iteration, and for tests that don't have dependences we can let the full test set run ?

test/k6/src/tests/orders_email.js
```diff
@@ -64,169 +61,119 @@ export function setup() {
...
   });
 
-  addErrorCount(success);
-
-  if (!success) {
-    stopIterationOnFail(success);
-  }
-
-  return selfLink;
+  return response.headers["Location"];
```
acn-sbuad, line 86: is it explicit that this is the self link for the order? Might be easier to follow understand the return value if it is named despite it not being used further?

SandGrainOne: Reverted the change.

test/k6/src/tests/orders_sms.js
SandGrainOne: Added an else, but keeping the POST test on top for both scenario.

test/k6/src/tests/orders_email.js
```diff
@@ -1,54 +1,51 @@
...
-  },
+    // Checks rate should be 100%. Raise error if any check has failed.
+    checks: ['rate>=1']
+  }
 };
 
 export function setup() {
-  var token = setupToken.getAltinnTokenForOrg(scopes);
-  var sendersReference = uuidv4();
+  const token = setupToken.getAltinnTokenForOrg("altinn:serviceowner/notifications.create");
```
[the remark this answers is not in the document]
SandGrainOne: Yeah, looking into the lifecycle more in detail I see it was wrong to move more stuff into setup. Should move stuff out instead. I'll revert my changes.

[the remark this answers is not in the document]
SandGrainOne: Correction. global values like scope belong in setup, but I'm still reverting to be consistent. Can look into lifecycle in more detail later.

acn-sbuad (approved):
🎈

PR #572, opened 2024-07-05
Add condition checks to tests

SandGrainOne:
## Description
Expand the automated tests to include condition checks when making an email notification order.

## Related Issue(s)
- #549 
- #562

## Verification
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

### DIFF
--- a/test/k6/src/config.js
+++ b/test/k6/src/config.js
@@ -1,5 +1,5 @@
 // Baseurls for platform
-export var baseUrls = {
+const baseUrls = {
   at21: "at21.altinn.cloud",
   at22: "at22.altinn.cloud",
   at23: "at23.altinn.cloud",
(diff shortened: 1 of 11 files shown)


test/k6/src/config.js
```diff
@@ -44,6 +44,9 @@ export var notifications = {
   },
   notifications_sms: function (orderId) {
     return `https://platform.${baseUrl}/notifications/api/v1/orders/${orderId}/notifications/sms/`;
+  },
+  conditionCheck: function(conditionMet) {
+    return `http://altinn-notifications.default.svc.cluster.local/notifications/api/v1/tests/sendcondition?conditionMet=${conditionMet}`;
```
acn-sbuad, line 49: this would require changes if we ever move the deployment out of the default namespace in AKS. Would recommend that we rather use the endpoint exposed via APIM so we don't have a hidden dependency here

acn-sbuad: How about we make the full url an input param in the test?

SandGrainOne: We can also generate a url based on environment like we do with other urls.

SandGrainOne: Adding subscription key as a separate value for now. With some dynamic changes to the url to change the condition result I find it best to generate the url in code.

Issue #616, opened 2024-09-24
Regression testing of Notifications with Profile configured to use local KRR cop as data source


olebhansen:
Blocked by https://github.com/Altinn/altinn-profile/issues/176

Once our new KRR replica is available, we should refactor notifications to use this as a datasource in applicable cases (when the recipient is a person identified by SSN).

### Tasks
- [ ] Check if Altinn 2 is using the KRR dates: "updated" and "verified" and how if so.
- [x] Ensure that Notifications handles the newly declared "nullability" in properties UserId, NationalIdentityNumber, MobileNumber, and Email
- [x] Test order creation that involves users having null-values for UserId, NationalIdentityNumber, MobileNumber, and/or Email

### Acceptance criteris:
- [x] Notification uses Altinn3 KRR replica when this is the only datasource needed (persons by SSN)

### Things to discuss
- **Threat modelling**
Discuss possible security issues
- **Documentation**
- [ ] Do we need to update documentation in altinn docs?
- [ ] Do we need Internal documentation (confluence)?
- [x] We do not need to update documentation or write internal documentation
- **External tests**
- [ ] Should new functionality be Regression tested?
- [ ] Should new functionality be included in Usecase tests?
- [x] Do not need to update automated tests
- **Acceptance criteria**
Does the issue have defined acceptance criteria?

PR #644, opened 2024-11-12
Prepare for nullability of UserId in Profile API - user contact point lookups

hggutvik:
### Current state
The Profile API is in the process of phasing out setting actual data in the UserId field when looking up user contact points (i.e. when POSTing to _profile/api/v1/users/contactpoint/lookup_ and _profile/api/v1/users/contactpoint/availability_)

Notifications uses a domain model ([UserContactPoints](https://github.com/Altinn/altinn-notifications/blob/main/src/Altinn.Notifications.Core/Models/ContactPoints/UserContactPoints.cs)) with no nullable-declared properties, which also serves as the deserialization model when looking up user contact points in the [ProfileClient](https://github.com/Altinn/altinn-notifications/blob/main/src/Altinn.Notifications.Integrations/Profile/ProfileClient.cs#L50). This leads to a runtime exception if/when the relevant  Profile API endpoints start returning `userId: null`, as the domain model expects an integer value. The other properties of the ([UserContactPoints](https://github.com/Altinn/altinn-notifications/blob/main/src/Altinn.Notifications.Core/Models/ContactPoints/UserContactPoints.cs)) domain model are capable of receiving null values from the Profile API, as they are initialized to `string.Empty`.

To avoid this breaking change in the `UserId `property, the Profile API is currently returning `userId: 0` for user contact points lookups.

### Changes in this PR
Add a DTO / API model `UserContactPointsDTO` that represents/reflects the current/soon-to-be API contract from the Profile API when looking up user contact points. The DTO explicitly declares nullable properties. An explicit mapper (for DTO -> domain model) is defined (currently implemented as [an extension method on the DTO class](https://github.com/Altinn/altinn-notifications/pull/644/files#diff-724a02ba3034d5995742e3868b029cf163160f9e75b1503a936bbc3b69cfadf3)) and [used by the ProfileClient](https://github.com/Altinn/altinn-notifications/pull/644/files#diff-e7bfa1a77951fc70932caeb58ed79ff245507752d74bcf85e9ae55a1d6d5bf04) when returning the data to the caller (the Core [ContactPointsService](https://github.com/Altinn/altinn-notifications/blob/main/src/Altinn.Notifications.Core/Services/ContactPointService.cs)).

Using this separate API / DTO model, as compared to modifying the domain model [UserContactPoints](https://github.com/Altinn/altinn-notifications/blob/main/src/Altinn.Notifications.Core/Models/ContactPoints/UserContactPoints.cs), has the advantage of decoupling the current direct binding between the Profile API interface and our core / domain model in this application. 

## Related Issue(s)
- #616 

## Verification
- [x] All tests run green

## Documentation


src/Altinn.Notifications.Integrations/Profile/Models/UserContactPointsDTO.cs
SandGrainOne, line 1: DTO is capitalized, maybe use Dto instead, like the class name.

SandGrainOne (approved):
Very good.

Issue #688, opened 2025-01-10
Implement token generation for authentication in Notification API


Ahmed-Ghanam:
### Description

We want the Notifications API to securely access external APIs, such as the Register API, by using an access token. The access token will be used to authenticate and authorize requests to these APIs, ensuring secure and efficient communication.

The Notifications API should generate tokens dynamically when required and manage their lifecycle, including expiration and renewal.

#### Example
A sample request with an access token:
```
POST /register/api/v1/parties/nameslookup HTTP/1.1
PlatformAccessToken: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...
Content-Type: application/json

{
    "ssn": [
        "12345678901"
    ]
}

```

### Additional Information

- The access token should be a JSON Web Token (JWT) to ensure compatibility with standard authentication mechanisms.
- Token generation must include support for claims and scopes, specifying what the token can access and for how long.
- Follow industry standards for token encryption and signing (e.g., RSA256).

### Refinement
- Tokens should be cached for reuse to minimize unnecessary generation.
- Ensure the system can handle token renewal seamlessly, avoiding downtime or failed requests.

### Tasks
- [x] Implement token generation service
- [x] Unit/integration tests

### Acceptance Criteria
- [ ] Access tokens are successfully generated, validated, and cached.
- [ ] Tokens are included in API requests for secure communication.

Ahmed-Ghanam:
This ticket builds upon the implementation provided in #545.

Ahmed-Ghanam:
The values.yaml file has been updated in this [pull request](https://dev.azure.com/brreg/altinn-studio-ops/_git/altinn-studio-ops/pullrequest/1174).

PR #690, opened 2025-01-10
Implement Access Token creation logic

Ahmed-Ghanam:
## Description
Implement the necessary logic to create an access token.

## Related Issue(s)
- #688 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.Notifications.Integrations/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.Notifications.Integrations/Extensions/ServiceCollectionExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using Altinn.ApiClients.Maskinporten.Extensions;
 using Altinn.ApiClients.Maskinporten.Services;
+using Altinn.Common.AccessTokenClient.Services;
 using Altinn.Common.PEP.Clients;
 using Altinn.Common.PEP.Implementation;
 using Altinn.Common.PEP.Interfaces;
@@ -13,6 +14,7 @@ using Altinn.Notifications.Integrations.Kafka.Producers;
 using Altinn.Notifications.Integrations.Register;
 using Altinn.Notifications.Integrations.SendCondition;
 
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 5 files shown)


src/Altinn.Notifications.Integrations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -70,14 +72,16 @@ public static void AddAuthorizationService(this IServiceCollection services, ICo
         services.Configure<Common.PEP.Configuration.PlatformSettings>(config.GetSection("PlatformSettings"));
         services.AddHttpClient<AuthorizationApiClient>();
         services.AddSingleton<IPDP, PDPAppSI>();
+        services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```
SandGrainOne, line 75: I'm not sure if this is necessary. I don't think AccessTokenGenerator has a dependency to HttpContext.

Ahmed-Ghanam: I removed it. I see the same configuration is a part of the program.cs

src/Altinn.Notifications.Core/Integrations/IRegisterClient.cs
```diff
@@ -27,5 +27,5 @@ public interface IRegisterClient
     /// A task that represents the asynchronous operation. 
     /// The task result contains a list of <see cref="PartyDetails"/> representing the details of the specified individuals and organizations.
     /// </returns>
-    Task<List<PartyDetails>> GetPartyDetails(List<string> organizationNumbers, List<string> socialSecurityNumbers);
+    Task<List<PartyDetails>> GetPartyDetails(List<string>? organizationNumbers = null, List<string>? socialSecurityNumbers = null);
```
SandGrainOne, line 30: Adding null as default value was probably unnecessary.

My opinion on default values are that we should avoid them. Use them only to retain backwards compatibility when adding a parameter. That way you don't need to update all call points. Especially useful in a library. When the call points are few it is better to update the call points.

src/Altinn.Notifications.Integrations/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -13,6 +14,7 @@
 using Altinn.Notifications.Integrations.Register;
 using Altinn.Notifications.Integrations.SendCondition;
 
+using Microsoft.AspNetCore.Http;
```
SandGrainOne, line 17: Unused using

Ahmed-Ghanam: Removed.

SandGrainOne (approved):

Issue #1084, opened 2025-10-09
Using Task.Delay in Kafka consumer integration tests is causing tests to take too long


eskebab:
### Description

By utilising Task.Delay in Kafka consumer integration tests to initialize infrastructure causes unnecessary wait.
We should use the EventuallyAsync method introduced by @Ahmed-Ghanam to poll for results at a given polling interval, to get the desired result as fast as possible. This ensures we don't wait longer than necessary for consumption and production.

### Additional Information

Look in consumer test classes for the use of EventuallyAsync to understand how this can be used with mock invocations.

PR #1109, opened 2025-10-21
Feature/1084 remove task delay

eskebab:
Remove task.delay(10000) for integration tests involving Kafka consumers

## Description
By utilising Task.Delay in Kafka consumer integration tests to initialize infrastructure causes unnecessary wait.
We should use the EventuallyAsync method introduced by @Ahmed-Ghanam to poll for results at a given polling interval, to get the desired result as fast as possible. This ensures we don't wait longer than necessary for consumption and production.

## Related Issue(s)
- #1084 

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **Tests**
  * Improved integration test reliability by replacing fixed delays with condition-based polling mechanisms that wait for specific outcomes to occur
  * Enhanced test execution flow to verify results more robustly using polling intervals instead of static wait times


Ahmed-Ghanam (approved):
**👍 Great work!**  
I've updated the tests to start the consumer before publishing messages to the topic, ensuring they better reflect real-world scenarios. Previously, publishing the message before starting the consumer caused the tests to fail silently when run locally.

Issue #1298, opened 2026-02-13
Skip external calls that have no added value on order acceptance


olebhansen:
During order acceptance (POST to /future/orders), there are some validations that consumes resources, but add no value.

E.g. for org., when having the mandated official contact info, there is no need to parse through all optional contacts (see #1295 )

Change the code to diffrentiate between accepting and processing the order, and minimize the work during acceptance.

This must not introduce any functional change as to what orders are accepted/recected or who gets notified.
This should result in a minor speed-up for order acceptance.

PR #1347, opened 2026-02-27
Feature 1298 skip auth on order acceptance

eskebab:
During order acceptance (POST to /future/orders), there are some validations that consumes resources, but add no value.

E.g. for org., when having the mandated official contact info, there is no need to parse through all optional contacts (see https://github.com/Altinn/altinn-notifications/issues/1295 )

Change the code to diffrentiate between accepting and processing the order, and minimize the work during acceptance.

This must not introduce any functional change as to what orders are accepted/recected or who gets notified.
This should result in a minor speed-up for order acceptance.

## Related Issue(s)
- #1298 

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Added OrderPhase (NewOrder, Processing) to represent order lifecycle phases.

* **Public API**
  * Contact-point operations (email, SMS, email+SMS, preferred) now accept an optional OrderPhase parameter to indicate creation vs processing.

* **Tests**
  * Added and updated unit and integration tests for NewOrder vs Processing behavior.
  * Added test utilities (spy and factory) to capture and verify phase-aware contact-point calls.

### DIFF
--- a/components/api/test/Altinn.Notifications.Tests/Notifications.Core/TestingServices/ContactPointServiceTests.cs
+++ b/components/api/test/Altinn.Notifications.Tests/Notifications.Core/TestingServices/ContactPointServiceTests.cs
@@ -41,7 +41,7 @@ namespace Altinn.Notifications.Tests.Notifications.Core.TestingServices
 
             // Act & Assert
             var exception = await Assert.ThrowsAsync<PlatformDependencyException>(
-                async () => await service.AddEmailContactPoints(recipientsToEnrich, null, It.IsAny<OrderLifecycleStage>()));
+                async () => await service.AddEmailContactPoints(recipientsToEnrich, null, OrderLifecycleStage.Registration));
 
             Assert.Equal("ProfileClient", exception.DependencyName);
             Assert.Equal("GetUserContactPoints", exception.Operation);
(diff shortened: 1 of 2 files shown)


components/api/src/Altinn.Notifications.Core/Services/ContactPointService.cs
```diff
@@ -574,18 +586,24 @@ private async Task<List<OrganizationContactPoints>> GetOfficialOrganizationConta
...
         List<string> organizationNumbers,
-        string sanitizedResourceId)
+        string sanitizedResourceId,
+        OrderPhase orderPhase)
     {
-        List<OrganizationContactPoints> allUserContactPoints = await GetUserRegisteredContactPoints(organizationNumbers, sanitizedResourceId);
+        List<OrganizationContactPoints> authorizedUserContactPoints;
 
-        if (allUserContactPoints.Count == 0)
+        if (orderPhase == OrderPhase.Processing)
```
sigridbra: If we are to skip all calls to userRegisteredContactPoints, why not have the if-check on  line 551?

sigridbra: This would change todays behaviour slightly. 

GetOfficialOrganizationContactPoints gets official contact points that are mandatory for the organization. However, it is only mandatory to have one, that can be either email or sms. 
For example, some orgs may have only email registered. In that case, if a notification is ordered to channel SMS, GetOfficialOrganizationContactPoints  will return no contact points. When we skip  GetUserRegisteredContactPoints as well, the result is no contact points and the order will return with 422.

Issue #1364, opened 2026-03-03
Create Wolverine handler for SMS delivery status updates


martivj:

martivj:
**Manual test report — SMS delivery report handling via ASB**

Tested the ASB solution for SMS delivery reports both locally and in AT22. Each test uses the same setup with an Order in `Processing`, SMS notification in `Accepted`, and a known gateway reference (UUID). For local testing we used PurpleExplorer to observe retry attempts on the queue; in AT22 we used peek on the Service Bus instance in the Azure portal. Delivery reports were simulated via a curl request directly to the webhook, XML-formatted with the correct credentials per environment.

**Local:**

**Happy path (Delivered):**
Input: delivery report with State `DELIVRD`, matching gateway reference
Result: Order updated to `Completed`, SMS notification to `Delivered`, new status feed entry :white_check_mark:

**Happy path 2 (Failed_BarredReceiver):**
Input: delivery report with State `BARRED`, matching gateway reference
Result: Order updated to `Completed`, SMS notification to `Failed_BarredReceiver`, new status feed entry :white_check_mark:
Confirmed correct mapping sent to publisher

**Not found:**
Input: delivery report with State `DELIVRD`, unknown gateway reference
Result: 7 attempts over ~5 min -> dead delivery report with `RETRY_THRESHOLD_EXCEEDED` :white_check_mark:

**Expired:**
Input: delivery report with State `DELIVRD`, matching gateway reference
Result: Order still `Processing`, SMS notification still `Accepted`, dead delivery report with `NOTIFICATION_EXPIRED` without retry attempts :white_check_mark:
Expected, since expired status on order/notification is handled by the cron job

**Database disabled mid retry processing:**
Input: delivery report with State `DELIVRD`, unknown gateway reference
Result: `ObjectDisposedException`, message on DLQ without retry :white_check_mark:
Unexpected exception type, but confirmed that default handling outside defined chains is DLQ without retry (expected)

**AT22:**

Happy path: same result as local :white_check_mark:
Not found: same result as local :white_check_mark:
Expired: not tested — expected result would be dead delivery report without retry + cron job updates status
Difficult to simulate infrastructure failures; we only depend on the database (no auth/profile lookups inside the handler)

martivj:
We want to officially test the implementation in TT02 to see the full flow of `/future/orders` in Bruno -> SMS delivery report handling all the way to the statusfeed. 

[...]

PR #1426, opened 2026-03-24
Implement SMS delivery report transfer with ASB

martivj:
## Related Issue(s)
- #1364

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Added SMS delivery report handling through Azure Service Bus, enabling processing of SMS delivery status updates from messaging gateways.
  * Configurable SMS delivery report queue settings with independent retry policies.

* **Tests**
  * Added comprehensive SMS delivery report handler integration tests covering success and failure scenarios.
  * Extended test utilities with SMS notification and delivery report helpers.

### DIFF
--- a/components/api/src/Altinn.Notifications.Integrations/Wolverine/Handlers/SmsDeliveryReportHandler.cs
+++ b/components/api/src/Altinn.Notifications.Integrations/Wolverine/Handlers/SmsDeliveryReportHandler.cs
@@ -30,7 +30,7 @@ public static class SmsDeliveryReportHandler
         }
 
         if (!Enum.TryParse<Core.Enums.SmsNotificationResultType>(command.SendResult, ignoreCase: true, out var sendResult) ||
-            !Enum.IsDefined(typeof(Core.Enums.SmsNotificationResultType), sendResult))
+            !Enum.IsDefined(sendResult))
         {
             logger.LogError("Received SMS delivery report with unrecognized SendResult: {SendResult}", command.SendResult);
             throw new InvalidDeliveryReportException($"Received SMS delivery report with unrecognized SendResult: '{command.SendResult}'.");


components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+/// <summary>
+/// Handles SMS delivery status updates received from the Azure Service Bus queue.
+/// </summary>
+[ExcludeFromCodeCoverage]
+public static class SmsDeliveryReportHandler
+{
+    /// <summary>
+    /// Gets or sets the Wolverine settings used for configuring error handling policies.
+    /// </summary>
+    public static WolverineSettings Settings { get; set; } = null!;
```
Ahmed-Ghanam: Static mutable state is shared across the entire application lifetime. This means:
* Thread safety issues: Any thread can set/overwrite it at any time
* Test isolation breaks: Tests run in the same process share the same static state, causing flaky tests
* Hidden coupling: Classes that depend on it don't declare that dependency — it's invisible

Settings should be injected (via constructor or method parameter), not set through a global static property. This makes the dependency:
* Explicit (visible in the signature)
* Testable (easy to mock/substitute)
* Lifecycle-managed (by the DI container)

The fact that this pattern has leaked into the `main` branch is not a reason to follow the same pattern forward because it weakens the code.

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// <summary>
+    /// Gets the reason code that indicates the retry threshold has been exceeded.
+    /// </summary>
+    public static string RetryExceededReason => "RETRY_THRESHOLD_EXCEEDED";
+
+    /// <summary>
+    /// Gets the reason code indicating that a notification has expired.
+    /// </summary>
+    public static string NotificationExpiredReason => "NOTIFICATION_EXPIRED";
```
Ahmed-Ghanam: Since these are just fixed string literals, and used internally, a property adds unnecessary overhead. const is more idiomatic:
``` c#
private const string NotificationExpiredReason = "NOTIFICATION_EXPIRED";
private const string RetryExceededReason = "RETRY_THRESHOLD_EXCEEDED";
```

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// <summary>
+    /// Configures error handling for the SMS delivery report queue handler.
+    /// </summary>
+    public static void Configure(HandlerChain chain)
+    {
+        var policy = Settings.SmsDeliveryReportQueuePolicy;
```
Ahmed-Ghanam: Settings should be injected to remove this coupling

components/api/src/Altinn.Notifications.Integrations/Wolverine/FailureActionsExtensions.cs
```diff
@@ -18,34 +20,57 @@ public static class FailureActionsExtensions
...
     {
         return failureActions.CustomAction(
             async (runtime, envelope, exception) =>
             {
-                var command = (EmailDeliveryReportCommand)envelope.Envelope!.Message!;
-                var eventGridEvent = EventGridEvent.Parse(command.Message.Body);
+                string payload = channel switch
+                {
+                    DeliveryReportChannel.AzureCommunicationServices => ExtractEmailPayload(envelope.Envelope!.Message!),
+                    DeliveryReportChannel.LinkMobility => ExtractSmsPayload(envelope.Envelope!.Message!),
```
Ahmed-Ghanam: Using the `null-forgiving operator` follows some general rules, and I think it acceptable only when:
* A framework/DI container guarantees initialization (e.g., [Required] model binding)
* You've just checked for null yourself and the compiler can't infer it
* It's in test code where brevity matters more

The `null-forgiving operator` tells the compiler "I guarantee this isn't null" — but if either `Envelope` or `Message` is null at runtime, you get a `NullReferenceException` with no context about what was null or why. Explicit null checks with meaningful errors is eetter approach:

```c#
 async (runtime, envelope, exception) =>
 {
     var innerEnvelope = envelope.Envelope 
         ?? throw new InvalidOperationException("Wolverine envelope is null.");
     
     var message = innerEnvelope.Message 
         ?? throw new InvalidOperationException("Envelope message is null.");
 
     string payload = channel switch
     {
         DeliveryReportChannel.AzureCommunicationServices => ExtractEmailPayload(message),
         DeliveryReportChannel.LinkMobility => ExtractSmsPayload(message),
         _ => throw new InvalidOperationException($"No payload extractor for channel {channel}")
     };
     // ...
 }
```

components/api/src/Altinn.Notifications.Integrations/Wolverine/FailureActionsExtensions.cs
```diff
@@ -18,34 +20,57 @@ public static class FailureActionsExtensions
...
+                    Reason = reason,
+                    Message = exception.Message
+                };
+
+                await deadDeliveryReportService.InsertAsync(deadDeliveryReport);
             },
             "Save Dead Delivery Report");
     }
+
+    private static string ExtractEmailPayload(object message)
```
Ahmed-Ghanam, line 72: Personal preferences: Adding XML doc is always a good practice.

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        chain
+            .OnException<NotificationNotFoundException>()
+            .RetryWithCooldown(policy.GetCooldownDelays())
+            .Then.ScheduleRetry(policy.GetScheduleDelays())
+            .Then.SaveDeadDeliveryReport(RetryExceededReason, Core.Enums.DeliveryReportChannel.LinkMobility);
+
+        chain
+            .OnException<NotificationExpiredException>()
+            .SaveDeadDeliveryReport(NotificationExpiredReason, Core.Enums.DeliveryReportChannel.LinkMobility);
```
Ahmed-Ghanam: In `FailureActionsExtensions`, the switch expression throws an `InvalidOperationException` when it encounters an unknown channel:

```csharp
_ => throw new InvalidOperationException($"No payload extractor registered for channel {channel}")
```

This `InvalidOperationException` is raised inside `SaveDeadDeliveryReport`, which itself is the configured failure action for the `NotificationNotFoundException` chain. If Wolverine re‑evaluates the handler chains when this failure action throws, the message would match Chain 1 again and trigger retries and `MoveToErrorQueue` even though the issue is actually an unknown channel, not a transient failure.

To avoid this incorrect behavior, we can throw a more specific exception type that does *not* belong to any handler chain:

```csharp
_ => throw new NotSupportedException($"No payload extractor registered for channel {channel}")
```

Alternatively, an application‑specific exception makes the intent even clearer:

```csharp
_ => throw new InvalidDataException($"No payload extractor registered for channel {channel}")
```

Neither `InvalidDataException` nor `NotSupportedException` is included in any of the three configured chains, so they will correctly fall back to Wolverine’s default behavior of moving the message to the error queue — which is the desired outcome.

Note: Same applies for `EmailDeliveryReportHandler`

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
```diff
@@ -0,0 +1,264 @@
...
+{
+    private readonly IntegrationTestContainersFixture _fixture = fixture;
+
+    /// <summary>
+    /// Tests the happy path: a delivery report with status "Delivered" is received
+    /// for an SMS notification that exists in the database with a matching gatewayReference.
+    /// The handler should update the notification status to "Delivered".
+    /// </summary>
+    [Fact]
+    public async Task SmsDeliveryReport_WhenNotificationExists_UpdatesStatusToDelivered()
```
Ahmed-Ghanam: XML documentation does not belong in tests because they are designed for public API consumers — they power IntelliSense, generate NuGet documentation, and describe contracts. Test methods have no external consumers. A well-named test method is self-documenting

I think we need to find this rule and remove it

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
```diff
@@ -0,0 +1,264 @@
...
+    {
+        var factory = new IntegrationTestWebApplicationFactory(_fixture).Initialize();
+
+        await using (factory)
+        {
+            // Arrange - Create notification and set status to Accepted with a gatewayReference
+            // (simulates the SMS service having successfully sent via Link Mobility)
+            var (_, notification) = await PostgreUtil.PopulateDBWithOrderAndSmsNotification(factory);
```
Ahmed-Ghanam, line 36: **Rule of thumb**: Every test must clean up the data it used for the test after the test is completed.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
```diff
@@ -0,0 +1,264 @@
...
+                async () =>
+                {
+                    var id = await PostgreUtil.GetDeadDeliveryReportByGatewayReference(
+                         _fixture.PostgresConnectionString,
+                         unmatchedGatewayReference);
+                    return id.HasValue;
+                },
+                maxAttempts: 40,
+                delayMs: 500);
+            Assert.True(deadReportFound, "Dead delivery report should be saved after retries are exhausted");
```
Ahmed-Ghanam, line 108: Don't we need to verify `reason`, `channel`, `attemptCount`, or `resolved`?

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
```diff
@@ -0,0 +1,264 @@
...
+
+namespace Altinn.Notifications.IntegrationTestsASB.Tests;
+
+/// <summary>
+/// Integration tests for the SMS delivery report Wolverine handler.
+/// Sends SmsDeliveryReportCommand messages to the ASB queue via Wolverine (simulating the SMS service)
+/// and verifies end-to-end processing through the handler pipeline.
+/// </summary>
+[Collection(nameof(IntegrationTestContainersCollection))]
+public class SmsDeliveryReportHandlerTests(IntegrationTestContainersFixture fixture)
```
Ahmed-Ghanam, line 23: Invalid `SendResult` string - `Enum.Parse<SmsNotificationResultType>(command.SendResult)` would throw `ArgumentException`, which is not in any error chain and goes straight to DLQ with no retries. This is worth an explicit test.

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
martivj: This is now extracted out to in dedicated `IHandlerPolicy` classes for both delivery report handlers.

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
martivj: These are now `private const string` in dedicated `IHandlerPolicy` classes.

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
martivj: We now use the `IHandlerPolicy` class pattern to add policies outside the handlers themselves. Dedicated policy classes have been added, namespaced in the subdirectory `Wolverine/Policies`.

components/api/src/Altinn.Notifications.Integrations/Wolverine/FailureActionsExtensions.cs
martivj: This has been added now, but with `InvalidDataException` to avoid retry logic.

components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsDeliveryReportHandler.cs
martivj: We now use `NotSupportedException` to avoid retries.

components/api/src/Altinn.Notifications.Integrations/Wolverine/FailureActionsExtensions.cs
martivj: XML docs have been added for both extractor methods.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
martivj: We now have a GlobalSuppressions.cs file to ignore `SA1600:Elements should be documented`. The XML docs in both delivery reports' integration test files have been removed.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
martivj: The `CleanupAsync` on lines 84-101 in the `IntegrationTestWebApplicationFactory` already deletes everything from all relevant tables. I agree that we could make it more explicit per test, but we have a catch-all in place already.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
martivj: We have now introduced a `DeadDeliveryReportRow` record to verify the entire row in a type-safe manner.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsDeliveryReportHandlerTests.cs
martivj: A new integration test has been added.

components/sms-service/test/Altinn.Notifications.Sms.IntegrationTestsASB/Infrastructure/IntegrationTestWebApplicationFactory.cs
```diff
@@ -36,6 +45,6 @@ protected override void ConfigureComponentServices(IConfiguration configuration,
     /// <inheritdoc/>
     protected override async Task DrainQueuesAsync()
     {
-        await DrainDeadLetterQueuesAsync(Fixture.ServiceBusConnectionString, "smoke-test");
+        await DrainDeadLetterQueuesAsync(Fixture.ServiceBusConnectionString, WolverineSettings.SmsDeliveryReportQueueName);
     }
```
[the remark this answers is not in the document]
martivj: If the queue name was null, we would have problems elsewhere too. This guard just clutters the code in my opinion.

components/api/src/Altinn.Notifications/Program.cs
```diff
@@ -228,10 +228,7 @@ void ConfigureServices(IServiceCollection services, IConfiguration config)
     services.AddCoreServices(config);
     services.AddAuthorizationService(config);
     services.AddKafkaServices(config);
-    if (config.GetValue<bool>("WolverineSettings:EnableWolverine"))
-    {
-        services.AddWolverineServices(config, builder.Environment);
-    }
+    services.AddWolverineServices(config, builder.Environment);
```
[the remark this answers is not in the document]
martivj: That is out of scope for this PR. We have this issue to switch patterns over to the new DI setup: https://github.com/Altinn/altinn-notifications/issues/1452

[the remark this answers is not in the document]
martivj: Actually, we can move it to be gated in the private method still, done in the latest commit.

Ahmed-Ghanam (approved):
Approved this after approval from Espen. I did not go through the implementation to avoid duplicate code review.

PR #1467, opened 2026-04-14
Implement email send operation result via ASB

Ahmed-Ghanam:
## Description
Extends the Azure Service Bus migration to cover the email send result flow and SMS delivery reports, completing the ASB transport path alongside the existing completing the ASB transport path alongside the existing Kafka path.

## Related Issue(s)
- #1455

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * End-to-end email send-result flow: publish and consume email send outcomes so final statuses propagate across services with resilient retry and error routing.

* **Configuration**
  * New toggles, queue names and retry policies to enable/disable and validate send-result publisher/listener; runtime validation added.

* **Tests**
  * Extensive unit and integration tests for success, retry/backoff, dead-letter, unknown-result handling, and JSON serialization.

* **Chores**
  * Reordered background services, updated queue draining order, and large DB migration adding notification-processing routines.


components/email-service/src/Altinn.Notifications.Email.Integrations/Producers/EmailSendResultProducer.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        _producer = producer;
+        _topicName = topicName;
+    }
+
+    /// <inheritdoc/>
+    public async Task DispatchAsync(SendOperationResult result)
+    {
+        bool success = await _producer.ProduceAsync(_topicName, result.Serialize());
+        if (!success)
```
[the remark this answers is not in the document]
Ahmed-Ghanam: We are going to replace Kafka path with ASB path so we do not need to improve the validation rules there to avoid breaking changes.

Ahmed-Ghanam:
I’ve updated the base implementation of the Kafka consumer, which is why there may be some duplication in this PR. There’s nothing that needs to be addressed at the moment.

components/api/src/Altinn.Notifications.Integrations/Wolverine/Policies/EmailSendResultHandlerPolicy.cs
```diff
@@ -0,0 +1,50 @@
...
+            .Or<ServiceBusException>()
+            .Or<TaskCanceledException>()
+            .RetryWithCooldown(policy.GetCooldownDelays())
+            .Then.ScheduleRetry(policy.GetScheduleDelays())
+            .Then.MoveToErrorQueue();
+
+        chain
+            .OnException<ArgumentException>()
+            .SaveDeadDeliveryReport(_unrecognizedSendResultReason, DeliveryReportChannel.AzureCommunicationServices);
+    }
```
martivj, line 49: Cool solution for the unknown send result cases! I assume we could do exception handling in the email component when we try to map the unknown result type, but with your solution we are guaranteed to store down the delivery report so they don't get lost, nice work.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/EmailSendResultHandlerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        var factory = new IntegrationTestWebApplicationFactory(_fixture)
+            .ReplaceService(_ => mockService.Object)
+            .Initialize();
+
+        await using (factory)
+        {
+            var policy = factory.WolverineSettings!.EmailSendResultQueuePolicy;
+            int expectedAttempts = 1 + policy.CooldownDelaysMs.Length + policy.ScheduleDelaysMs.Length;
+
+            string queueName = factory.WolverineSettings.EmailSendResultQueueName;
```
martivj, line 82: Very nice pattern to extract `expectedAttempts` dynamically. We should consider making an issue to apply the same pattern in all Wolverine handler tests across all components (most tests hardcode the number of attempts, but if the config ever changes the test would break, so this pattern is way more stable).

Maybe make a new issue for this improvement across the other test projects?

components/email-service/src/Altinn.Notifications.Email/appsettings.json
```diff
@@ -56,6 +56,12 @@
     "EmailStatusCheckQueuePolicy": {
       "CooldownDelaysMs": [ 1000, 5000, 10000 ],
       "ScheduleDelaysMs": [ 60000, 120000, 120000 ]
+    },
+    "EnableEmailSendResultPublisher": false,
+    "EmailSendResultQueueName": "altinn.notifications.email.send.result",
+    "EmailSendResultQueuePolicy": {
```
martivj, line 62: The `EmailSendResultQueuePolicy` should be removed. Policies only apply to listeners, and the email component only contains a publisher for this queue.

components/email-service/src/Altinn.Notifications.Email.Integrations/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -107,4 +108,31 @@ private static void RegisterEmailStatusCheckDispatcher(IServiceCollection servic
...
         }
     }
+
+    /// <summary>
+    /// Registers the appropriate <see cref="IEmailSendResultDispatcher"/> implementation
+    /// based on Wolverine configuration, selecting either the ASB or Kafka transport path.
+    /// </summary>
+    private static void RegisterEmailSendResultDispatcher(IServiceCollection services, WolverineSettings wolverineSettings, KafkaSettings kafkaSettings)
+    {
+        services.RemoveAll<IEmailSendResultDispatcher>();
```
martivj, line 118: Can the `RemoveAll<IEmailSendResultDispatcher>` on line 118 be considered dead code? We don't register the dispatcher anywhere else prior to this call.

The same applies on line 90 for the `RemoveAll<IEmailStatusCheckDispatcher>`, we can remove this line if we are sure this function is the only DI registration call point.

components/email-service/src/Altinn.Notifications.Email.Integrations/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -107,4 +108,31 @@ private static void RegisterEmailStatusCheckDispatcher(IServiceCollection servic
...
+        if (wolverineSettings.EnableWolverine && wolverineSettings.EnableEmailSendResultPublisher)
+        {
+            if (string.IsNullOrWhiteSpace(wolverineSettings.EmailSendResultQueueName))
+            {
+                throw new InvalidOperationException(
+                    $"{nameof(WolverineSettings.EmailSendResultQueueName)} must be configured when {nameof(WolverineSettings.EnableEmailSendResultPublisher)} is enabled.");
+            }
+
+            services.AddSingleton<IEmailSendResultDispatcher, EmailSendResultPublisher>();
+        }
```
martivj, line 129: Nice pattern with `InvalidOperationException`, we should update the other components to use the same setup for `ServiceCollectionExtensions`, either on this branch or in a new issue/PR.

components/email-service/src/Altinn.Notifications.Email.Integrations/Configuration/WolverineSettings.cs
```diff
@@ -45,7 +41,23 @@ public class WolverineSettings : WolverineSettingsBase
...
+    /// <summary>
+    /// ASB queue name for publishing email send results.
+    /// Produced by this email service and consumed by the Notifications API.
+    /// </summary>
+    public string EmailSendResultQueueName { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Retry policy for the email send result queue.
+    /// </summary>
+    public QueueRetryPolicy EmailSendResultQueuePolicy { get; set; } = new();
```
martivj, line 62: Remove the unused policy field (see `appsettings.json` comment).

components/email-service/src/Altinn.Notifications.Email.Integrations/Producers/EmailSendResultProducer.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// </param>
+    /// <param name="topicName">
+    /// The name of the Kafka topic where the terminal result will be published.
+    /// </param>
+    public EmailSendResultProducer(ICommonProducer producer, string topicName)
+    {
+        _producer = producer;
+        _topicName = topicName;
+    }
+
```
martivj, line 30: Personal preference: I like to be consistent with using primary constructors.

components/email-service/src/Altinn.Notifications.Email.Integrations/Publishers/EmailSendResultPublisher.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// Initializes a new instance of the <see cref="EmailSendResultPublisher"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">
+    /// The service provider used to resolve a scoped <see cref="IMessageBus"/> instance for each dispatch.
+    /// </param>
+    public EmailSendResultPublisher(IServiceProvider serviceProvider)
+    {
+        _serviceProvider = serviceProvider;
+    }
+
```
martivj, line 30: Personal preference: I like to be consistent with using primary constructors.

components/email-service/test/Altinn.Notifications.Email.IntegrationTestsASB/appsettings.integrationtest.json
```diff
@@ -30,6 +30,12 @@
     "EmailStatusCheckQueuePolicy": {
       "CooldownDelaysMs": [ 100, 100, 100 ],
       "ScheduleDelaysMs": [ 500, 500, 500 ]
+    },
+    "EnableEmailSendResultPublisher": false,
+    "EmailSendResultQueueName": "altinn.notifications.email.send.result",
+    "EmailSendResultQueuePolicy": {
```
martivj, line 36: Remove the `EmailSendResultQueuePolicy` (ref previous comments)

components/api/src/Altinn.Notifications.Persistence/Migration/v0.76/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,2348 @@
+-- This script is autogenerated from the tool DbTools. Do not edit manually.
```
martivj, line 1: Was this committed by mistake? I see no updates in the functions and procedures migrations.

martivj (changes requested):
Nice work. We need to remove the retry policy definitions in the email component where we only have a publisher for the `EmailSendResult` queue, and maybe make some new issues to align other components with your new patterns.

components/api/src/Altinn.Notifications.Integrations/Wolverine/Policies/EmailSendResultHandlerPolicy.cs
Ahmed-Ghanam: You're right that we could catch the mapping error in the email component, but there's a key difference: at that point we're still on the email service side and would have to decide what to do with the unknown result — likely log and drop it. By letting it propagate as an ArgumentException to the API handler, Wolverine's policy intercepts it and calls `SaveDeadDeliveryReport`, which means the raw payload is persisted and can be inspected or replayed later. Nothing gets silently swallowed.

It also keeps the email component clean — it doesn't need to know about dead delivery storage or recovery strategies. That concern lives entirely in the API's handler policy where it belongs.

components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/EmailSendResultHandlerTests.cs
Ahmed-Ghanam: [Use dynamic expectedAttempts in all Wolverine handler integration tests](https://github.com/Altinn/altinn-notifications/issues/1489)

components/email-service/src/Altinn.Notifications.Email.Integrations/Configuration/ServiceCollectionExtensions.cs
Ahmed-Ghanam: [Apply fail-fast InvalidOperationException pattern to all ServiceCollectionExtensions Wolverine registrations](https://github.com/Altinn/altinn-notifications/issues/1490)

components/email-service/src/Altinn.Notifications.Email.Integrations/Publishers/EmailSendResultPublisher.cs
Ahmed-Ghanam: Since this PR already uses classic constructors with validation logic in the body, switching to primary constructors would actually mean removing those guards or moving them elsewhere. That's a regression, not an improvement.

components/email-service/src/Altinn.Notifications.Email.Integrations/Producers/EmailSendResultProducer.cs
Ahmed-Ghanam: Since this PR already uses classic constructors with validation logic in the body, switching to primary constructors would actually mean removing those guards or moving them elsewhere. That's a regression, not an improvement.

components/api/src/Altinn.Notifications.Persistence/Migration/v0.76/01-functions-and-procedures.sql
Ahmed-Ghanam: Nope. this is coming from this PR https://github.com/Altinn/altinn-notifications/pull/1434

martivj (approved):

PR #1473, opened 2026-04-15
1439 update k6 tests to use correct instant endpoints

### LINKED ISSUE #1439: Update k6 tests to use correct instant endpoints
## Description

The current k6 test setup at `k6/src/tests/orders-v2.js` uses an `orders_sms_instant_v2` endpoint variable (at `k6/src/config.js`) that targets the old deprecated SMS-only instant endpoint `future/orders/instant`. We should update the tests to use the channel specific `future/orders/instant/sms` and `future/orders/instant/email` with their new request payloads (`timeToLiveInSeconds` is only relevant for SMS etc.), per the [instant orders docs](https://docs.altinn.studio/nb/notifications/reference/openapi/#/Instant%20Orders).

## Acceptance Criteria

- [x] The k6 tests uses channel specific endpoints, both email and SMS
- [x] The tests use the appropriate required request payload structure

ugedal1:
## Description
  - Fixed the SMS instant notification endpoint to use the channel-specific     
  route (/instant/sms)
  - Added support for email instant notification orders, including test data,   
  API client function, threshold labels
  - Updated the SMS instant order payload structure

## Related Issue(s)
- #1439 

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **Tests**
  * Expanded API test coverage to include instant email and an older instant-SMS flow, with new scenarios, labels and stronger posting and shipment-status validations.
* **Chores**
  * Added new test fixtures and updated test data shapes to support the new instant variants.
* **Chores**
  * CI workflows updated to pass an email-recipient environment flag to the API test runs.

### DIFF
--- a/components/api/test/k6/src/api/notifications/v2.js
+++ b/components/api/test/k6/src/api/notifications/v2.js
@@ -11,22 +11,22 @@ export function postNotificationOrderV2(serializedOrder, token, label) {
   return http.post(endpoint, serializedOrder, params);
 }
 
-export function postSmsInstantNotificationOrderRequest(data, token) {
+export function postSmsInstantNotificationOrderRequest(data, token, label) {
   const endpoint = config.notifications.orders_sms_instant_v2;
 
   const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
-  params.tags = { name: "POST SMS instant notification order request" };
+  params.tags = { name: label };
 
   const smsOrderRequest = data;
 
   return http.post(endpoint, smsOrderRequest, params);
 }
 
-export function postEmailInstantNotificationOrderRequest(data, token) {
+export function postEmailInstantNotificationOrderRequest(data, token, label) {
   const endpoint = config.notifications.orders_email_instant_v2;
 
   const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
-  params.tags = { name: "POST Email instant notification order request" };
+  params.tags = { name: label };
 
   const emailOrderRequest = data;
 
--- a/components/api/test/k6/src/api/notifications/v2.js
+++ b/components/api/test/k6/src/api/notifications/v2.js
@@ -17,9 +17,16 @@ export function postSmsInstantNotificationOrderRequest(data, token, label) {
   const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
   params.tags = { name: label };
 
-  const smsOrderRequest = data;
+  return http.post(endpoint, data, params);
+}
 
-  return http.post(endpoint, smsOrderRequest, params);
+export function postSmsInstantNotificationOrderRequestOld(data, token, label) {
+  const endpoint = config.notifications.orders_sms_instant_old_v2;
+
+  const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
+  params.tags = { name: label };
+
+  return http.post(endpoint, data, params);
 }
 
 export function postEmailInstantNotificationOrderRequest(data, token, label) {
@@ -28,9 +35,7 @@ export function postEmailInstantNotificationOrderRequest(data, token, label) {
   const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
   params.tags = { name: label };
 
-  const emailOrderRequest = data;
-
-  return http.post(endpoint, emailOrderRequest, params);
+  return http.post(endpoint, data, params);
 }
 
 export function getShipment(orderId, token, label) {
--- a/components/api/test/k6/src/api/notifications/v2.js
+++ b/components/api/test/k6/src/api/notifications/v2.js
@@ -22,6 +22,17 @@ export function postSmsInstantNotificationOrderRequest(data, token) {
   return http.post(endpoint, smsOrderRequest, params);
 }
 
+export function postEmailInstantNotificationOrderRequest(data, token) {
+  const endpoint = config.notifications.orders_email_instant_v2;
+
+  const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
+  params.tags = { name: "POST Email instant notification order request" };
+
+  const emailOrderRequest = data;
+
+  return http.post(endpoint, emailOrderRequest, params);
+}
+
 export function getShipment(orderId, token, label) {
   const params = apiHelpers.buildHeaderWithBearer(token);
   params.tags = { name: label };
--- a/components/api/test/k6/src/config.js
+++ b/components/api/test/k6/src/config.js
@@ -3,30 +3,27 @@ import { environment } from "./shared/variables.js";
 
 // Base URLs for the Altinn platform across different environments.
 const baseUrls = {
-  prod: "altinn.no",
-  tt02: "tt02.altinn.no",
-  yt01: "yt01.altinn.cloud",
-  at22: "at22.altinn.cloud",
-  at23: "at23.altinn.cloud",
-  at24: "at24.altinn.cloud",
+    prod: "altinn.no",
+    tt02: "tt02.altinn.no",
+    yt01: "yt01.altinn.cloud",
+    at22: "at22.altinn.cloud",
+    at23: "at23.altinn.cloud",
+    at24: "at24.altinn.cloud"
 };
 
 // Base URLs for Maskinporten authentication service in different environments.
 const maskinportenBaseUrls = {
-  prod: "https://maskinporten.no/",
-  tt02: "https://test.maskinporten.no/",
+    prod: "https://maskinporten.no/",
+    tt02: "https://test.maskinporten.no/"
 };
 
 if (!environment) {
-  stopIterationOnFail("Environment variable 'altinn_env' is not set", false);
+    stopIterationOnFail("Environment variable 'altinn_env' is not set", false);
 }
 
 const baseUrl = baseUrls[environment];
 if (!baseUrl) {
-  stopIterationOnFail(
-    `Invalid value for environment variable 'altinn_env': '${environment}'.`,
-    false,
-  );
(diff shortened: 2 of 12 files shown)


components/api/test/k6/src/data/orders/order-v2-sms-instant.json
```diff
@@ -1,14 +1,12 @@
...
-      }
+  "recipientSms": {
+    "phoneNumber": "string",
+    "timeToLiveInSeconds": 60,
+    "smsSettings": {
+      "sender": "Altinn-test",
+      "body": "This is a plain test body for testing SMS instant services"
     }
   }
 }
```
Ahmed-Ghanam, line 12: The `/future/orders/instant` endpoint is still functional and actively in use. We also don’t have any concrete plans to remove it, so the existing test needs to be retained in my opinion.

components/api/test/k6/src/data/orders/order-v2-email-instant.json
```diff
@@ -0,0 +1,13 @@
+{
+  "idempotencyId": "string",
+  "sendersReference": "string",
+  "recipientEmail": {
+    "emailAddress": "string",
+    "emailSettings": {
+      "subject": "Automated instant email test from Altinn",
+      "body": "This is a plain test body for testing Email instant services",
+      "senderEmailAddress": "noreply@altinn.no",
```
Ahmed-Ghanam, line 9: Hard‑coding the sender email address is brittle, as we do not use a single, consistent sender address across all environments

components/api/test/k6/src/config.js
```diff
@@ -43,7 +43,8 @@ export const notifications = {
 
     orders_v2: `https://platform.${baseUrl}/notifications/api/v1/future/orders/`,
 
-    orders_sms_instant_v2: `https://platform.${baseUrl}/notifications/api/v1/future/orders/instant`,
+    orders_sms_instant_v2: `https://platform.${baseUrl}/notifications/api/v1/future/orders/instant/sms`,
```
Ahmed-Ghanam, line 46: Since we need to keep the existing test against `future/orders/instant`, the address should be retained.

components/api/test/k6/src/tests/orders-v2.js
```diff
@@ -32,12 +32,13 @@ import * as futureOrdersApi from "../api/notifications/v2.js";
...
     setEmptyThresholds, get_email_shipment, get_sms_shipment,
-    get_sms_instant_shipment, get_status_feed
+    get_sms_instant_shipment, get_email_instant_shipment, get_status_feed,
+    post_email_instant_order_v2
 } from "./threshold-labels.js";
 import { scopes } from "../shared/variables.js";
 import { getEmailRecipient, getSmsRecipient } from "../shared/functions.js";
 
-const labels = [post_email_order_v2, post_sms_order_v2, post_sms_instant_order_v2, get_email_shipment, get_sms_shipment, get_status_feed];
+const labels = [post_email_order_v2, post_sms_order_v2, post_sms_instant_order_v2, post_email_instant_order_v2, get_email_shipment, get_sms_shipment, get_status_feed];
```
Ahmed-Ghanam, line 41: Given that the `summaryTrendStats` configuration includes `p(95)`, `p(99)`, `p(99.5)`, and `p(99.9)`, adding the missing labels `get_sms_instant_shipment` and `get_email_instant_shipment` would be consistent with the intent of the test setup

components/api/test/k6/src/tests/orders-v2.js
```diff
@@ -102,12 +106,19 @@ export function setup() {
...
+
+    const emailOrderInstantRequest = {
+        ...emailOrderInstantRequestJson,
+        idempotencyId: uuidv4(),
+        sendersReference,
+        recipientEmail: {
+            ...emailOrderInstantRequestJson.recipientEmail,
+            emailAddress: emailRecipient
         }
     };
```
Ahmed-Ghanam, line 123: The instant endpoints differ fundamentally from the regular ones — `recipientSms` and `recipientEmail` are the only supported recipient types here. When running these endpoints manually against non‑YT01 environments, both instant requests will silently submit null as the phone number or email address, resulting in an API validation error rather than a clear misconfiguration message.

components/api/test/k6/src/api/notifications/v2.js
```diff
@@ -11,17 +11,28 @@ export function postNotificationOrderV2(serializedOrder, token, label) {
...
 
-export function postSmsInstantNotificationOrderRequest(data, token) {
+export function postSmsInstantNotificationOrderRequest(data, token, label) {
   const endpoint = config.notifications.orders_sms_instant_v2;
 
   const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
-  params.tags = { name: "POST SMS instant notification order request" };
+  params.tags = { name: label };
 
   const smsOrderRequest = data;
```
Ahmed-Ghanam, line 20: This `const smsOrderRequest = data` adds noise with no benefit. `data` can be passed directly.

components/api/test/k6/src/api/notifications/v2.js
```diff
@@ -11,17 +11,28 @@ export function postNotificationOrderV2(serializedOrder, token, label) {
...
   return http.post(endpoint, smsOrderRequest, params);
 }
 
+export function postEmailInstantNotificationOrderRequest(data, token, label) {
+  const endpoint = config.notifications.orders_email_instant_v2;
+
+  const params = apiHelpers.buildHeaderWithBearerAndContentType(token);
+  params.tags = { name: label };
+
+  const emailOrderRequest = data;
```
Ahmed-Ghanam, line 31: This `const emailOrderRequest = data` adds noise with no benefit. `data` can be passed directly.

Ahmed-Ghanam (changes requested):

components/api/test/k6/src/tests/orders-v2.js
ugedal1: Will add checks earlier in the setup function to make it clear if it is misconfigured.

ugedal1:
- Added back setup with old endpoint, all files,variables, functions etc. are marked with old to seperate it from the new endpoints.
- Added early in the setup function which stops the configuration if sms and emailrecipient has not been checked. Removed previous dead code which is no longer accessable because of this.
- removed hardcoded sender adress from instant email json

components/api/test/k6/src/tests/orders-v2.js
Ahmed-Ghanam, line 72: The `stopIterationOnFail `calls k6’s `fail()`, which throws and aborts `setup()`. Because of this, lines where the phone number and email are mutated are only reached if none of the guards trigger, so execution itself is safe.

The issue is a behavioral change. Previously, `emailRecipient` was optional—if it was null, `recipientEmail` was simply not set. With this change, both `emailRecipient` and `smsRecipient` are now effectively required.
In environments where `getSmsRecipient()` or `getEmailRecipient()` return `null` (for example, non‑YT01 environments where the environment variables are not set), the test suite now fails immediately during `setup()`. This is a breaking change for anyone running the tests against AT22, AT23, TT02, or production without both variables defined.
As a result, the command‑line usage comment at the top of the file is now misleading, as these parameters are no longer optional in practice. We need to make them optional

Ahmed-Ghanam (changes requested):

components/api/test/k6/src/tests/orders-v2.js
ugedal1: Have updated the tests(except performance since those are run in yt) to use the AUTOMATED_EMAIL which is found as a github secret. This worked for both at environments and tt02, have not tested the other environments yet.

components/api/test/k6/src/tests/orders-v2.js
```diff
@@ -74,49 +95,62 @@ export function setup() {
...
-    }
 
-    if (ninRecipient == null) {
-        // unset recipientPerson object when no national identity number is provided
-        emailOrderRequestJson.recipient["recipientPerson"] = undefined;
-    }
-    else {
+
+    //can only have one active recipient
+    if (ninRecipient != null && !emailRecipient) {
```
Ahmed-Ghanam: Decouple the instant-order guard from the regular email order logic. `emailRecipient` can be mandatory for instant email while still being optional for NIN-based regular email orders.

components/api/test/k6/src/tests/orders-v2.js
Ahmed-Ghanam: The mandatory `emailRecipient` guard executes before the `ninRecipient` logic:

```
// Guard ensures emailRecipient is ALWAYS truthy past this point
if (!emailRecipient) {
    stopIterationOnFail("emailRecipient is required...", false);
}
```

```
// ...later...
if (ninRecipient != null && !emailRecipient) { // !emailRecipient is ALWAYS false
    emailOrderRequestJson.recipient.recipientPerson.nationalIdentityNumber = ninRecipient;
} else {
    emailOrderRequestJson.recipient["recipientPerson"] = undefined; // always executes
}
```

Because of this ordering, `recipientPerson` is always set to undefined. As a result, the `ninRecipient` environment variable is accepted but silently ignored. This means that NIN‑based recipient lookup is never exercised, even though that was the original purpose of `ninRecipient` in the regular email order test.

The core problem is that logic intended for instant email orders is now affecting the behavior of regular email orders. The instant‑order guard should be decoupled from the regular email order logic. `emailRecipient` can be mandatory for instant email orders while remaining optional for regular email orders that use `ninRecipient`. The same for `smsRecipient`

Ahmed-Ghanam (changes requested):

components/api/test/k6/src/tests/orders-v2.js
ugedal1: decoupled the code so normal order requests use nin if given and instant always use email.

Ahmed-Ghanam (approved):

PR #1479, opened 2026-04-16
Add daily email metrics

### LINKED ISSUE #1460: Extend stats export to cover emails
Emails will also be part of invoicing. We need to export statistics similar to SMS. The size does not need to be a parameter.

sigridbra:
## Related Issue(s)
- #1460 

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Added daily email metrics API and matching parquet download alongside existing SMS metrics.
  * Parquet export generalized to support both email and SMS metrics.

* **Tests**
  * Added integration and unit tests covering email metrics retrieval, parquet generation, and authorization scenarios.

* **Bug Fixes**
  * Authorization filter extended to validate API keys for the email metrics route.


components/api/test/Altinn.Notifications.IntegrationTests/Notifications.Persistence/MetricsRepositoryTests.cs
```diff
@@ -83,4 +83,29 @@ public async Task GetDailySmsMetrics()
...
+            .GetServices([typeof(IMetricsRepository)])
+            .First(i => i.GetType() == typeof(MetricsRepository));
+
+        var date = DateTime.UtcNow;
+
+        NotificationOrder order = await PostgreUtil.PopulateDBWithOrderAnd4Notifications(orgName, date.AddDays(-1));
+        _orderIdsToDelete.Add(order.Id);
+
+        // Act
+        var result = await sut.GetDailyEmailMetrics(date.Day, date.Month, date.Year, CancellationToken.None);
```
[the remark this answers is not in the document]
sigridbra: Yes, something is wrong with the setup so the notifications end up the wrong day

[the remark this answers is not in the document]
sigridbra: The error is somewhere in the PopulateDBWithOrderAnd4Notifications helper but I havent found the source

components/api/src/Altinn.Notifications.Persistence/Migration/v0.77/01-setup-index.sql
```diff
@@ -0,0 +1,3 @@
+-- To support efficient querying for metrics in emailnotifications
+CREATE INDEX IF NOT EXISTS notifications_emailnotifications_resulttime ON notifications.emailnotifications (resulttime);
```
sigridbra, line 2: I have done some analysis of this with Claude. We believe that this will be a significant improvement. Explain analyze shows that right now, the lookup in emailnotifications is the problem as it first finds 10 million rows before filtering out the relevant rows. 

If anything more were to be done Claude suggest adding a covering index to the orders table with just the columns we need as the lookup of every row uses a lot of memory because the rows are so large.

components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getemailmetrics.sql
```diff
@@ -0,0 +1,43 @@
...
+  start_date TIMESTAMPTZ;
+BEGIN
+  start_date = MAKE_TIMESTAMPTZ(year_input, month_input, day_input, 0, 0, 0, 'UTC');
+
+  RETURN QUERY
+  select 
+    -- references and correlation
+    email._id as email_id --unique recipient/"functional email"
+,   email.alternateid as shipmentid --unique per notification/reminder (but the same for the same notification/reminder to multiple recipients, e.g. to an organization where people with access to the resource have custom contact information)
+,   orders.notificationorder ->> 'SendersReference' as senders_reference --senders reference (not necessarily unique)
```
Ahmed-Ghanam: Since the `orders` table already has a dedicated `sendersreference` column, it would be preferable to read from that column directly. Extracting the value from JSONB is less efficient than accessing a native column

components/api/src/Altinn.Notifications.Persistence/Migration/v0.78/01-setup-index.sql
```diff
@@ -0,0 +1,3 @@
+-- To support efficient querying for metrics in emailnotifications
+CREATE INDEX IF NOT EXISTS notifications_emailnotifications_resulttime ON notifications.emailnotifications (resulttime);
```
Ahmed-Ghanam: The query in `notifications.get_email_metrics` consistently filters on `result NOT IN ('New', 'Sending', 'Succeeded')`.

The existing non‑partial index includes all rows, including those that will always be excluded by this predicate. As a result, PostgreSQL must scan index entries for these rows, fetch the corresponding heap pages, and then discard them during filtering.

A better approach would be to use a **partial index** that only includes rows the query can actually return. This would make the index smaller and faster to scan, while also reducing unnecessary heap access.

Example:
```sql
CREATE INDEX notifications_emailnotifications_resulttime_metrics
  ON notifications.emailnotifications (resulttime)
  INCLUDE (_id, alternateid, _orderid, result, operationid)
  WHERE result NOT IN (
    'New'::emailnotificationresulttype,
    'Sending'::emailnotificationresulttype,
    'Succeeded'::emailnotificationresulttype
  );
```

Using `INCLUDE` also makes the index covering, allowing PostgreSQL to satisfy the query entirely from the index and eliminate heap fetches for the `emailnotifications` side of the query.

components/api/src/Altinn.Notifications.Persistence/Migration/v0.77/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,2394 @@
...
+,   orders.notificationorder ->> 'SendersReference' as senders_reference --senders reference (not necessarily unique)
+,   orders.requestedsendtime --requested sending time (to determine when it is correct to invoice, if applicable. May differ slightly from actual sending time, so check in combination with status/gateway ref)
+,   orders.creatorname --orderer's maskinporten ID (the real service owner can be hidden by aggregation, e.g. correspondence)    
+,   orders.notificationorder ->> 'ResourceId' as resourceid --resourceid, can in combination with creatorname provide real service owner or granulation corresponding to service code etc.
+
+     -- operator status
+,   email.result::text as result -- status of the delivery (error, but with an operationid may mean that the message was attempted sent/tariffed, but for various reasons did not reach the user)
+,   email.operationid -- reference at ACS (a reference likely means that ACS bills for this � but not necessarily)
```
Ahmed-Ghanam: The generator has converted the character `-` to `�`. Can we remove it from where it is introduced?

components/api/src/Altinn.Notifications.Core/Models/Metrics/DailyEmailMetricsRecord.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// The sender's reference associated with the shipment.
+        /// </summary>
+        public string SendersReference { get; init; } = string.Empty;
+
+        /// <summary>
+        /// The requested send time for the email.
+        /// </summary>
+        public string RequestedSendTime { get; init; } = string.Empty;
```
Ahmed-Ghanam, line 26: **Hint:** The records are serialized to Parquet via `ParquetSerializer.SerializeAsync`. Storing a timestamp as a string means the Parquet column is typed as `BYTE_ARRAY/UTF8` instead of `INT64/TIMESTAMP`. Any analytics consumer (Spark, Pandas, Power BI) will see a string column and lose the ability to sort, filter, or aggregate by time natively.

components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getemailmetrics.sql
```diff
@@ -0,0 +1,43 @@
...
+DECLARE
+  start_date TIMESTAMPTZ;
+BEGIN
+  start_date = MAKE_TIMESTAMPTZ(year_input, month_input, day_input, 0, 0, 0, 'UTC');
+
+  RETURN QUERY
+  select 
+    -- references and correlation
+    email._id as email_id --unique recipient/"functional email"
+,   email.alternateid as shipmentid --unique per notification/reminder (but the same for the same notification/reminder to multiple recipients, e.g. to an organization where people with access to the resource have custom contact information)
```
Ahmed-Ghanam, line 22: **Hint:** The `shipmentId` returned by the `POST /future/orders` endpoint corresponds to the `alternateid` column in the `orders` table. The GitHub issue does not specify which identifier should be reported, so it may be worth double‑checking this

components/api/src/Altinn.Notifications/Authorization/MetricsApiKeyFilter.cs
```diff
@@ -22,7 +22,7 @@ public class MetricsApiKeyFilter(
         public async Task OnAuthorizationAsync(AuthorizationFilterContext context)
         {
             // Only apply to Metrics endpoints
-            if (!context.HttpContext.Request.Path.StartsWithSegments("/notifications/api/v1/metrics/sms"))
+            if (!context.HttpContext.Request.Path.StartsWithSegments("/notifications/api/v1/metrics/sms") && !context.HttpContext.Request.Path.StartsWithSegments("/notifications/api/v1/metrics/email"))
```
Ahmed-Ghanam, line 25: The current approach is fragile. The filter relies on a string‑based route check, even though it is only registered on two specific actions via `ServiceFilter`. If the route prefix ever changes, the check will silently stop working and all requests will pass through.

components/api/test/Altinn.Notifications.IntegrationTests/Notifications.Persistence/MetricsRepositoryTests.cs
```diff
@@ -83,4 +83,29 @@ public async Task GetDailySmsMetrics()
...
+
+        // Assert
+        Assert.InRange(result.Metrics.Count, 2, int.MaxValue);
+
+        var metrics = result.Metrics.FirstOrDefault(m => m.CreatorName == orgName);
+        Assert.NotNull(metrics);
+        Assert.Equal(orgName, metrics.CreatorName);
```
Ahmed-Ghanam, line 109: The `GetDailyEmailMetrics` barely verifies `CreatorName`. `EmailId`, `ShipmentId`, `Result`, `OperationId`, `RequestipmentId`, `Result`, `OperationId`, `RequestedSendTime` are not verified.

Ahmed-Ghanam (changes requested):

components/api/src/Altinn.Notifications.Persistence/Migration/v0.78/01-setup-index.sql
sigridbra: The covering index is for the table orders (which is wide due to jsonb), not emailnotifications. 

Regarding a partial index I'm not sure about the effectiveness of this as mondays data returned 124999 rows, but only 162 rows with result in ('New', 'Sending', 'Succeeded'). Meaning that a non-partial index and a partial index would give almost the same query result.

components/api/src/Altinn.Notifications.Core/Models/Metrics/DailyEmailMetricsRecord.cs
sigridbra: This is how it is done for sms metrics. I think they should be consistent.

components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getemailmetrics.sql
sigridbra: This is how it is done for sms metrics. I think they should be consistent.

components/api/src/Altinn.Notifications/Authorization/MetricsApiKeyFilter.cs
sigridbra: Route changes are unlikely as this would require coordinated deploy with finOps team.

components/api/test/Altinn.Notifications.IntegrationTests/Notifications.Persistence/MetricsRepositoryTests.cs
sigridbra: Added some more asserts

components/api/src/Altinn.Notifications.Persistence/Migration/v0.77/01-functions-and-procedures.sql
sigridbra: Fixed

components/api/src/Altinn.Notifications/Authorization/MetricsApiKeyFilter.cs
Ahmed-Ghanam: We can add a test that would fail if the controller route (or the filter path) is renamed. That would help us detect a mismatch early, before it silently disables the API key validation.

Ahmed-Ghanam (approved):
Good work 🎉

PR #1481, opened 2026-04-16
Patch latest dotnet image digests

hggutvik:
Patch to today's published digest for getting the fix for a remaining `musl-utils` (unexploitable) vulnerability.
Also removes the ignored CVE from trivyignore, as it has been fixed in the latest Docker version


hggutvik:
@coderabbitai pause

SandGrainOne (approved):

PR #1590, opened 2026-05-19
Enhancement 1525 specific exception types

### LINKED ISSUE #1525: Introduce specific exception types for send-result dead delivery report handling
## Background

Both `EmailSendResultHandler` and `SmsSendResultHandler` currently throw `ArgumentException` when `command.SendResult` cannot be parsed. The Wolverine handler policies catch `ArgumentException` to save a dead delivery report and discard the message (no retry).

The problem is that `ArgumentException` is too broad. There are two cases that should result in a dead delivery report:

1. The `SendResult` value on the incoming command is unrecognized and cannot be parsed.
2. `UpdateSendStatus` in `EmailNotificationRepository` / `SmsNotificationRepository` receives data it cannot process (e.g., notification not found or invalid state).

Both of these are non-retriable data errors that belong in the dead delivery report. However, any *other* `ArgumentException` thrown elsewhere in the handler chain should not be caught by this chain at all — Wolverine's default behaviour (move to DLQ) is the correct outcome for those.

## Proposal

Introduce more specific exception types for both cases, and update the policies to catch exactly those types in the dead-delivery-report chain. Two options:

**Option A — single shared type:** Use `InvalidDataException` (existing .NET BCL type) for both the unrecognized send-result case and the repository invalid-data case. The policy catches only `InvalidDataException`.

**Option B — dedicated types:** Introduce one or two dedicated exceptions (e.g., `UnrecognizedSendResultException` for the parse failure, and a separate type for the repository case), and chain them with `.Or<>()` in the policy.

In either case:
- The dead-delivery-report chain catches only the agreed specific type(s).
- Any unrelated `ArgumentException` will no longer be caught and will fall through to Wolverine's default DLQ path — which is the correct behaviour.

## Affected files

- `components/api/src/Altinn.Notifications.Integrations/Wolverine/EmailSendResultHandler.cs`
- `components/api/src/Altinn.Notifications.Integrations/Wolverine/SmsSendResultHandler.cs`
- `components/api/src/Altinn.Notifications.Integrations/Wolverine/Policies/EmailSendResultHandlerPolicy.cs`
- `components/api/src/Altinn.Notifications.Integrations/Wolverine/Policies/SmsSendResultHandlerPolicy.cs`
- `components/api/src/Altinn.Notifications.Persistence/Repository/EmailNotificationRepository.cs`
- `components/api/src/Altinn.Notifications.Persistence/Repository/SmsNotificationRepository.cs`

eskebab:
## Related Issue(s)
- #1525 

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  - Added explicit non-retriable error types and a new dead-delivery reason for invalid/missing notification identifiers.

* **Bug Fixes**
  - More precise error handling when send-result values are unrecognized or identifiers are missing; such cases now produce dedicated terminal reports.

* **Documentation**
  - Clarified method docs to explain identifier semantics and the new error condition.

* **Tests**
  - Expanded unit and integration tests for send-result parsing, identifier validation, and dead-delivery reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/altinn-notifications/pull/1590?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/api/src/Altinn.Notifications.Integrations/Wolverine/Policies/SmsSendResultHandlerPolicy.cs
```diff
@@ -57,7 +58,11 @@ public void Apply(IReadOnlyList<HandlerChain> chains, GenerationRules rules, ISe
             .SaveDeadDeliveryReport(_notificationExpiredReason, DeliveryReportChannel.LinkMobility);
 
         chain
-            .OnException<ArgumentException>()
+            .OnException<UnrecognizedSendResultException>()
             .SaveDeadDeliveryReport(_unrecognizedSendResultReason, DeliveryReportChannel.LinkMobility);
+
+        chain
+            .OnException<InvalidNotificationIdentifierException>()
+            .SaveDeadDeliveryReport(_invalidNotificationIdentifierReason, DeliveryReportChannel.LinkMobility);
```
martivj, line 66: The unit tests added in this PR are asserting handler logic for the unrecognized send result case to some degree, but to actually test things properly (and include the `InvalidNotificationIdentifierException` case) we should add two tests in the `IntegrationTestsASB`-project: 

- One for `UnrecognizedSendResultException` (bad SendResult value) and 
- One for `InvalidNotificationIdentifierException` (both identifiers null/empty), 

The tests should follow the same flow as the `NotificationExpiredException` policy chain test `SmsSendResult_WhenNotificationExpired_SavesDeadDeliveryReportWithoutRetry` in the `components/api/test/Altinn.Notifications.IntegrationTestsASB/Tests/SmsSendResultHandlerTests.cs` class, asserting the dead delivery report arrives in the database with the correct reason.

martivj (changes requested):
Implementation looks solid, but we are missing test for the actual policies (needs to be done through integration tests for wolverine to wire up correctly).

martivj (approved):

PR #1648, opened 2026-06-10
Fix flaky unit tests: parallel-safe metrics listeners + invariant test culture

olebhansen:
## Description

Fixes two sources of intermittent unit-test failures (observed as random CI failures unrelated to the changes under test, e.g. on #1646).

### 1. Metrics tests counted measurements from parallel tests

The `MeterListener` helpers in `EmailDeliveryReportHandlerTests`, `SmsDeliveryReportHandlerTests` and `DeliveryReportMetricsTests` enabled measurement events for **any** meter whose name matched `DeliveryReportMetrics.MeterName`. Test classes run in parallel and each creates its own `DeliveryReportMetrics` instance — and tests that start the application host via `WebApplicationFactory` register yet another — so a listener in one test could count measurements emitted elsewhere. This randomly broke both the "exactly one measurement" and the "zero measurements" assertions.

The listeners now match on the test's **own `Meter` instance** (exposed `internal` for tests via the existing `InternalsVisibleTo`). This keeps full test parallelism — no `[Collection]` serialization needed, so the suite stays fast.

### 2. Validator tests depended on the machine's UI culture

FluentValidation localizes its default messages from the current UI culture, so assertions like `'National Identity Number' must not be empty.` fail on machines with a non-English culture (nb-NO gives `... kan ikke være tom.`). A new `xunit.runner.json` pins the unit-test assembly to the **invariant culture**, making validation messages, formatting and parsing identical on every machine and on CI.

## Verification

- [x] Verified on a nb-NO machine (which reproduced the culture failure): **6 consecutive full-suite runs, 1069/1069 passing, ~1s per run**
- [x] `Altinn.Notifications.IntegrationTests` builds clean (uses the same listener pattern, fixed there too)
- [x] No production behaviour change — the only src change is an `internal` accessor on `DeliveryReportMetrics`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation for telemetry metrics to prevent interference from parallel test execution.
  * Enhanced meter matching logic in test listeners for more reliable metric verification.

* **Chores**
  * Added test runner configuration for consistent test environment settings.

### DIFF
--- a/components/api/src/Altinn.Notifications.Integrations/Telemetry/DeliveryReportMetrics.cs
+++ b/components/api/src/Altinn.Notifications.Integrations/Telemetry/DeliveryReportMetrics.cs
@@ -26,6 +26,13 @@ public sealed class DeliveryReportMetrics : IDisposable
     private readonly Counter<long> _deliveryReportStatusCounter;
 
     /// <summary>
+    /// Gets the <see cref="System.Diagnostics.Metrics.Meter"/> owned by this instance.
+    /// Exposed so tests can listen to this exact instance instead of matching on the meter name,
+    /// which would also pick up measurements from other instances in parallel tests.
+    /// </summary>
+    internal Meter Meter => _meter;
+
+    /// <summary>
     /// Initializes a new instance of the <see cref="DeliveryReportMetrics"/> class.
     /// </summary>
     public DeliveryReportMetrics()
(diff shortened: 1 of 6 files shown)


olebhansen:
Lukker denne som superseded av #1651.

Flaky-test-fiksen herfra (parallel-safe delivery-report metrics + invariant test culture) er allerede inkludert i #1651 — 5 av 6 filer er byte-identiske, og den siste (`Altinn.Notifications.Tests.csproj`) avviker kun ved at denne branchen har eldre pakkeversjoner. #1651 er rebaset på main, CI-grønn og mergeklar, så fiksen leveres via den i stedet.

Ingenting går tapt ved å lukke denne.

Issue #1740, opened 2026-07-06
Duplicate order processing fills the `orders.pastdue` dead-letter queue


martivj:
## Summary

During high-volume performance testing (of #1694, not yet merged), the `orders.pastdue` ASB queue's dead-letter queue filled up with:

```
Order {id} was not in Processing state; transaction rolled back.
```

Sampled DLQ messages were checked against the database and the referenced orders were already in `Processed` status — i.e. the order had already been processed successfully once, and a second, duplicate message for the same order later failed and was dead-lettered.

## Root cause

`StartProcessingPastDueOrders` resets *all* orders in a batch back to `Registered` when `PublishAsync` is cancelled mid-flight — including orders whose ASB message had already been sent successfully.

- `components/api/src/Altinn.Notifications.Core/Services/OrderProcessingService.cs:80-92`
  ```csharp
  catch (OperationCanceledException)
  {
      var ordersToReset = failedOrders ?? pastDueOrders;
      foreach (var order in ordersToReset)
          await _orderRepository.SetProcessingStatus(order.Id, OrderProcessingStatus.Registered);
      throw;
  }
  ```
- `PastDueOrderPublisher.SendAsync` rethrows `OperationCanceledException` instead of adding the order to its `failed` bag (`components/api/src/Altinn.Notifications.Integrations/Wolverine/Publishers/PastDueOrderPublisher.cs:76-80`). Because `Task.WhenAll` doesn't stop already-completed sibling tasks, several *other* orders in the same batch of up to 50 can have already been published to ASB before one order's send observes cancellation and the whole `PublishAsync` call throws.
- Since the exception escapes before `PublishAsync` returns, `failedOrders` is never assigned (stays `null`), so `ordersToReset = pastDueOrders` — **every** order in the batch gets reset to `Registered`, including already-published ones.
- Those orders get republished on the next poll, producing two ASB messages for the same order.
- The first message to complete transitions `Processing` → `Processed`/`Completed` via the guarded update at `components/api/src/Altinn.Notifications.Persistence/Repository/OrderRepository.cs:1020-1030` (`_advanceStatusFromProcessingSql`, requires `processedstatus = 'Processing'`).
- The second (duplicate) message finds 0 rows affected and throws `InvalidOperationException("Order {id} was not in Processing state; transaction rolled back.")`.
[...]

PR #1741, opened 2026-07-07
Guard against duplicate order processing

martivj:
### What changed

- `PastDueOrderPublisher` no longer lets cancellation escape as an exception from a per-order send — it's now treated like any other send failure. This means `PublishAsync` always returns a complete, accurate list of which orders were actually not published, even if cancellation fires mid-batch.
- Added `OrderRepository.ResetProcessingToRegistered`, replacing the direct `SetProcessingStatus(..., Registered)` calls. Its SQL only applies `WHERE processedstatus = 'Processing'`, so it can never regress an order that's already advanced past that state.
- Added `OrderProcessingService.ResetOrdersToRegistered`, which resets orders one at a time with a per-order try/catch, so a single failure doesn't stop the rest of the batch from being reset.

### Why this works

`StartProcessingPastDueOrders`'s cancellation handler used to fall back to resetting the *entire* claimed batch whenever it couldn't tell which orders had been published — because `PublishAsync` could throw before returning that information. Now it never throws, so the handler always has an accurate list and only resets orders it can prove weren't published. The guard on the reset SQL is a second line of defense: even if a reset were ever attempted against the wrong order, it's now a no-op instead of silently regressing a completed order back to `Registered`. Together, this removes the path that was causing the same order to be published — and processed — twice.

### Tests

- New/updated tests in `PastDueOrderPublisherTests` and `OrderProcessingServiceTests` covering the fixed cancellation handling.
- New integration tests confirming the reset is a no-op once an order has advanced past `Processing`.

## Related Issue(s)
- #1740

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order processing recovery by rolling back in-progress orders to a registered state, enabling them to be safely re-claimed and retried.
  * Cancellation during sending and while waiting for capacity now results in per-order failure reporting, rather than interrupting the whole batch.
  * Past-due order processing continues for remaining items when an individual order is canceled or fails, and queued cancellations are handled without throwing.

### DIFF
--- a/components/api/src/Altinn.Notifications.Integrations/Wolverine/Publishers/PastDueOrderPublisher.cs
+++ b/components/api/src/Altinn.Notifications.Integrations/Wolverine/Publishers/PastDueOrderPublisher.cs
@@ -86,9 +86,10 @@ public class PastDueOrderPublisher(
             await messageBus.SendAsync(new ProcessPastDueOrderCommand { Order = order });
             return null;
         }
-        catch (OperationCanceledException)
+        catch (OperationCanceledException ex)
         {
             _logger.LogInformation(
+                ex,
                 "PastDueOrderPublisher cancelled before publishing order {OrderId}; reporting as unpublished.",
                 order.Id);
             return order;
--- a/components/api/src/Altinn.Notifications.Integrations/Wolverine/Publishers/PastDueOrderPublisher.cs
+++ b/components/api/src/Altinn.Notifications.Integrations/Wolverine/Publishers/PastDueOrderPublisher.cs
@@ -42,9 +42,21 @@ public class PastDueOrderPublisher(
         var failed = new ConcurrentBag<NotificationOrder>();
         using var semaphore = new SemaphoreSlim(_publishConcurrency);
 
+        // None of the per-order work below is allowed to throw — including cancellation while
+        // still queued on the semaphore — so Task.WhenAll always returns normally with a `failed`
+        // list that is a complete and accurate record of every order that was not published.
+        // Callers depend on that guarantee to safely retry only the orders that need it.
         await Task.WhenAll(orders.Select(async order =>
         {
-            await semaphore.WaitAsync(cancellationToken);
+            try
+            {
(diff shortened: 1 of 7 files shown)


components/api/src/Altinn.Notifications.Integrations/Wolverine/Publishers/PastDueOrderPublisher.cs
```diff
@@ -42,9 +42,21 @@ public async Task<IReadOnlyList<NotificationOrder>> PublishAsync(
...
+        // None of the per-order work below is allowed to throw — including cancellation while
+        // still queued on the semaphore — so Task.WhenAll always returns normally with a `failed`
+        // list that is a complete and accurate record of every order that was not published.
+        // Callers depend on that guarantee to safely retry only the orders that need it.
         await Task.WhenAll(orders.Select(async order =>
         {
-            await semaphore.WaitAsync(cancellationToken);
+            try
+            {
+                await semaphore.WaitAsync(cancellationToken);
```
SandGrainOne, line 53: If we don't want this to be cancelled we could also send in "None". Does the AI have any input on this?

martivj: When cancellation is signaled, we want to add every order that hasn't been processed yet to the "failed" list so we can reset to back to `Registered` quickly. So I would argue we need the token here.

Claude on this, for reference:
> If we pass None, a queued order just waits for its turn behind the concurrency limit and only fails once it reaches SendAsync, instead of bailing out the instant cancellation fires. That matters for two reasons: it makes PublishAsync slower to return during shutdown (exactly when we want it to unwind fast), and it means a queued order could get stuck behind a slow/hung in-flight send that isn't respecting cancellation itself. Passing the real token here is what lets us report back immediately and accurately which orders weren't published, which is the whole point of this fix.

SandGrainOne (approved):

PR #1720, opened 2026-06-30
1687 notification search by organization number

ugedal1:
## Description
Adds support for searching notifications by organization number in the dashboard API. The implementation follows the same structure as the existing. NIN-based notification search, adding a new endpoint, request model, validator, service method, repository query, and SQL function — all mirroring the NIN pattern.

## Related Issue(s)
- #1687 

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Summary by CodeRabbit

* **New Features**
  * Added dashboard notification lookup by organization number, including a new request model, validation, service/repository support, and a new API endpoint.
  * Extended dashboard delivery-attempt responses to include an optional organization number.
* **Bug Fixes**
  * Standardized dashboard date-range handling (default windows, UTC expectations, and ordering).
  * Improved NIN and organization-number request validation consistency.
* **Tests**
  * Added/updated unit and integration tests for org-number lookups, validation errors, empty results, cancellation, and response mapping.

PR #1593, opened 2026-05-20
feat: added dashboard endpoint to fetch notifications by NIN

ugedal1:
## Description
Added a new dashboard endpoint that returns all notifications (email and SMS) by personal recipient and when it appears as the acting recipient in an organisational context.

  Changes:
  - New SQL function `getnotificationsbynin` querying `emailnotifications` and  
  `smsnotifications`
  - New `DashboardRepository` + `IDashboardRepository` in the Persistence layer.
  - New `DashboardService` + `IDashboardService` in the Core layer.
  - New `DashboardController` exposing `GET
  /notifications/api/v1/dashboard/notifications/nin` (query params: `nin`,      
  `from`, `to`), guarded by the `CREATE_SCOPE_OR_PLATFORM_ACCESS` policy.       
  - New `DashboardNotification` model (id, order id, creator name, resource id, 
  sender's reference, requested send time, recipient org/NIN, channel, result,  
  result time).
  - Bruno collection entries for manual verification of the new endpoint.       

  ## Related Issue(s)
  - #1493

  ## Verification
  - [x] **Your** code builds clean without any errors or warnings
  - [x] Manual testing done (required)
  - [x] Relevant automated test added (if you find this hard, leave it and we'll
   help out)
  - [x] All tests run green

  ## Documentation
  - [ ] User documentation is updated with a separate linked PR in
  [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if       
  applicable)

## Summary by CodeRabbit

* **New Features**
  * Dashboard API to fetch notifications by recipient ID with optional date range (defaults to last 7 days); includes validation of date range and secured access.
  * Backend service, repository and data model added to support the dashboard endpoint and return notification metadata, channel, result and timestamps.

* **Database / Migrations**
  * New/updated SQL routines to consolidate email/SMS queries and power dashboard lookups.

* **Tests**
  * Added API request definitions for the new dashboard endpoint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/altinn-notifications/pull/1593?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1782, opened 2026-07-22
1780 dashboard endpoint to get notifications by email

### LINKED ISSUE #1780: Dashboard endpoint to get notifications by email
### Description

Currently, notifications can be looked up by NIN or org number via the dashboard endpoints. Add a new endpoint that does the same lookup with a given email.

- Same response structure/models as the existing NIN/orgnr endpoints
- Same underlying notification fetch logic, just a different lookup key (email instead of NIN/orgnr)

ugedal1:
Adds a new dashboard endpoint for looking up notifications by the recipient's email address, following the same structure as the existing NIN and organization number lookups (#1593, #1720).

- Added `notifications.get_notifications_by_email` SQL function. Unlike the NIN/org-number functions, it only queries `emailnotifications`and returns both `recipientnin` and `recipientorgno` columns since either may be set depending on whether the notification was sent in a personal or organizational context.
- Added repository implementation, reusing the existing shared helper code.
- Added request model and validator, using the same validations logic as previous implementations.
- Added `DashboardController` endpoint: `GET notifications/api/v1/future/dashboard/recipients/notifications/email`.
- Added new tests for repository, serve and controller functions.      

## Summary by CodeRabbit

* **New Features**
  * Added the ability to retrieve dashboard email notifications by recipient email address.
  * Supports optional date-range filtering, case-insensitive email matching, and defaults to the previous seven days when dates are omitted.
  * Validates email addresses and date ranges, with clear responses for invalid requests and canceled operations.
* **Tests**
  * Added comprehensive coverage for filtering, validation, empty results, cancellation, and recipient identity mapping.


components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getnotificationsbyemail.sql
```diff
@@ -0,0 +1,67 @@
...
+)
+LANGUAGE sql
+STABLE
+PARALLEL SAFE
+AS $$
+    SELECT
+        o.alternateid AS shipmentid,
+        o.sendersreference,
+        o.creatorname,
+        o.type AS notificationtype,
```
Ahmed-Ghanam: The `type` column is defined as an `enum`, see `public.notificationordertype`, and it is generally considered good practice to cast it explicitly to `TEXT`, as this makes the intent clear to anyone reading the query. Therefore, I would write it as `o.type::text AS notificationtype`

components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getnotificationsbyemail.sql
```diff
@@ -0,0 +1,67 @@
...
+        e.toaddress AS address,
+        'email'::text AS channel,
+        e.result::text AS result,
+        e.resulttime
+    FROM notifications.emailnotifications e
+    JOIN notifications.orders o ON o._id = e._orderid
+    WHERE lower(e.toaddress) = lower(_email)
+      AND o.requestedsendtime >= _from_date
+      AND o.requestedsendtime <  _to_date
+    ORDER BY requestedsendtime DESC;
```
Ahmed-Ghanam: Replace `ORDER BY requestedsendtime DESC` with `ORDER BY o.requestedsendtime DESC`

components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getnotificationsbyemail.sql
```diff
@@ -0,0 +1,67 @@
...
+        o.requestedsendtime,
+        e.recipientnin,
+        e.recipientorgno,
+        e.toaddress AS address,
+        'email'::text AS channel,
+        e.result::text AS result,
+        e.resulttime
+    FROM notifications.emailnotifications e
+    JOIN notifications.orders o ON o._id = e._orderid
+    WHERE lower(e.toaddress) = lower(_email)
```
Ahmed-Ghanam, line 42: **FYI:** The `WHERE lower(e.toaddress) = lower(_email)` predicate will cause a full sequential scan of the table on every call because there is no index on `toaddress` at all. We can test and add an index when need.

components/api/src/Altinn.Notifications.Persistence/Migration/v0.94/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,3103 @@
+-- This script is autogenerated from the tool DbTools. Do not edit manually.
```
Ahmed-Ghanam, line 1: This version has already been deployed to the AT-22 environment. If it was used for testing, the changes must be rolled back in AT-22 before this feature is merged into the main branch. This will prevent deployment failures caused by the script attempting to re-execute previously applied changes.

components/api/src/Altinn.Notifications.Persistence/Repository/DashboardRepository.cs
```diff
@@ -38,6 +39,12 @@ public Task<List<DashboardNotification>> GetDashboardNotificationsByOrgNumberAsy
         return GetDashboardNotificationsAsync(_getNotificationsByOrgNo, recipientOrgNo, dateTimeFrom, dateTimeTo, cancellationToken);
     }
 
+    /// <inheritdoc/>
+    public Task<List<DashboardNotification>> GetDashboardNotificationsByEmailAsync(string email, DateTime? dateTimeFrom, DateTime? dateTimeTo, CancellationToken cancellationToken)
+    {
+        return GetDashboardNotificationsAsync(_getNotificationsByEmail, email, dateTimeFrom, dateTimeTo, cancellationToken);
+    }
+
     private async Task<List<DashboardNotification>> GetDashboardNotificationsAsync(
```
Ahmed-Ghanam, line 58: I recommend adding XML documentation to this function to ensure that its intent, inputs, outputs, and behavior are properly documented.

components/api/src/Altinn.Notifications/Models/Dashboard/NotificationsByEmailRequestExt.cs
```diff
@@ -0,0 +1,17 @@
...
+/// Request model for fetching notifications by email address.
+/// </summary>
+public class NotificationsByEmailRequestExt : DashboardNotificationRequestExt
+{
+    /// <summary>
+    /// The email address of the recipient.
+    /// </summary>
+    [BindRequired]
+    [FromHeader(Name = "Email")]
+    public required string Email { get; set; }
```
Ahmed-Ghanam, line 16: Email addresses are considered sensitive. HTTP headers are routinely logged by proxies, API gateways, and middleware — often without masking

components/api/src/Altinn.Notifications/Validators/Dashboard/NotificationsByEmailRequestValidator.cs
```diff
@@ -0,0 +1,22 @@
+using Altinn.Notifications.Models.Dashboard;
+using FluentValidation;
+
+namespace Altinn.Notifications.Validators.Dashboard;
+
+/// <summary>
+/// Validator for <see cref="NotificationsByEmailRequestExt"/>.
+/// </summary>
+internal sealed class NotificationsByEmailRequestValidator : AbstractValidator<NotificationsByEmailRequestExt>
```
Ahmed-Ghanam, line 9: We usually cover validators with tests as well.

components/api/src/Altinn.Notifications/Models/Dashboard/NotificationsByEmailRequestExt.cs
ugedal1: In previous endpoints we moved nin and orgnr to header for security reasons, is there a better place to append them?

Ahmed-Ghanam: I executed the following query in both Production and TT02 and was unable to identify any sensitive data stored in Application Insights:
```
requests
| where cloud_RoleName contains "platform-notifications" and url contains "dashboard"
| project customDimensions
```
no further action needed.

components/api/src/Altinn.Notifications.Persistence/Migration/FunctionsAndProcedures/getnotificationsbyemail.sql
ugedal1: Will test if a new index is needed after deploying to tt02. Will then add index to all relevant environments as needed.

components/api/test/Altinn.Notifications.Tests/Notifications/TestingControllers/DashboardControllerTests.cs
Ahmed-Ghanam, line 22: I could not find any integration tests covering the dashboard endpoints, similar to those implemented in `FutureOrdersControllerTests` under `Altinn.Notifications.IntegrationTests.Notifications.TestingControllers`.

I believe adding these tests is important to ensure adequate test coverage. However, their absence should not block the release of this feature once it is ready. You may either create a separate issue to address the test coverage or implement the tests as part of this feature.

ugedal1: Made seperate [issue](https://github.com/Altinn/altinn-notifications/issues/1811) to implement integration tests for dashboard endpoints.

